PR #5, opened 2015-10-23
Updates docker ee-repo with the new start and stop scripts.

kobalski:
Adds starts and stop scripts to docker image. Start script writes hazelcast PID to a file and stop scripts stops hazelcast instance started in docker container.

### DIFF
--- a/hazelcast-enterprise/server.sh
+++ b/hazelcast-enterprise/server.sh
@@ -2,7 +2,7 @@
 
 PRG="$0"
 PRGDIR=`dirname "$PRG"`
-HAZELCAST_HOME=`cd "$PRGDIR/.." >/dev/null; pwd`/hazelcast
+HAZELCAST_HOME=`cd "$PRGDIR/.." >/dev/null; pwd`/
 PID_FILE=$HAZELCAST_HOME/hazelcast_instance.pid
 
 if [ "x$MIN_HEAP_SIZE" != "x" ]; then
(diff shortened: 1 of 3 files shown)


hazelcast-enterprise/server.sh
```diff
@@ -0,0 +1,27 @@
...
+
+if [ "x$MIN_HEAP_SIZE" != "x" ]; then
+	JAVA_OPTS="$JAVA_OPTS -Xms${MIN_HEAP_SIZE}"
+fi
+
+if [ "x$MAX_HEAP_SIZE" != "x" ]; then
+	JAVA_OPTS="$JAVA_OPTS -Xmx${MAX_HEAP_SIZE}"
+fi
+
+export CLASSPATH=$HAZELCAST_HOME/hazelcast-enterprise-$HZ_VERSION/lib/hazelcast-enterprise-all-3.5.3.jar
```
eminn: I think we should remove the hardcoded version from here.

bilalyasar: probably version will be updated from release script,

eminn: Don't we already have $HZ_VERSION placeholder for this ?

kobalski: I should use $HZ_VERSION. Thanks now it is correct. Release script should only update $HZ_VERSION.

PR #19, opened 2017-02-17
New image for Openshift Enterprise

noctarius:
Merged the current status of the Openshift Enterprise (RHEL 7 based) image into upstream master and renamed the original hazelcast-openshift to hazelcast-openshift-origin to prevent confusion between origin and enterprise versions.


mesutcelik:
Hi @noctarius !

Do we really need your whole `.m2` structure? I see the PR contains all your local maven repository.

noctarius:
No it is "only" what maven needs to be kicked off. As you see there is no com/hazelcast packages in there. It is also not part of the resulting image, as it is deleted during the build after resolving the dependencies. It, however, is the better way to copy dependencies into the image as we don't need to have correct versions / urls of all of them inside the docker file. This would be very error prone for version updates.

emrahkocaman:
@noctarius Got your point but why do we need need ".m2" directory? `mvnw` will fetch related dependencies with maven dependency plugin.

hazelcast-openshift-rhel/Dockerfile
```diff
@@ -0,0 +1,93 @@
...
+ENV HZ_HOME /opt/hazelcast/
+ENV HZ_CP_MOUNT ${HZ_HOME}/external
+ENV JAVA_HOME /usr/lib/jvm/java
+ENV LANG en_US.utf8
+
+ENV USER_NAME=hazelcast
+ENV USER_UID=10001
+
+ARG HZ_VERSION=3.8-RC1
+ARG HZ_KUBE_VERSION=0.9.3
```
emrahkocaman: This should be updated as v1.0.0 is released. (https://github.com/hazelcast/hazelcast-kubernetes/releases/tag/v1.0.0)

hazelcast-openshift-rhel/mvnw/dependency-copy.xml
```diff
@@ -0,0 +1,42 @@
...
+      </exclusions>
+    </dependency>
+  </dependencies>
+  <repositories>
+    <repository>
+      <id>snapshot-repository</id>
+      <name>Maven2 Snapshot Repository</name>
+      <url>https://oss.sonatype.org/content/repositories/snapshots</url>
+    </repository>
+  </repositories>
```
emrahkocaman: As v1.0.0 is released, we don't need to add this repository.

noctarius:
No idea, this is what comes up when you create a new maven wrapper so I expect it to be needed. If not we can delete it.

hazelcast-openshift-rhel/Dockerfile
noctarius: Shit missed a bunch of commits, gonna fix it.

noctarius:
Ok tried it and it works without it, weird that this directory is actually created, whatever :) Thanks 👍 (I actually didn't like it myself).

Also fixed the wrong version of the kubernetes plugin, forgot to cherry-pick that commit.

noctarius:
Also added the README file with some basic documentation on how to use the template and create a cluster.

hazelcast-openshift-rhel/Dockerfile
```diff
@@ -0,0 +1,93 @@
...
+
+ENV HZ_HOME /opt/hazelcast/
+ENV HZ_CP_MOUNT ${HZ_HOME}/external
+ENV JAVA_HOME /usr/lib/jvm/java
+ENV LANG en_US.utf8
+
+ENV USER_NAME=hazelcast
+ENV USER_UID=10001
+
+ARG HZ_VERSION=3.8-RC1
```
mesutcelik: we can use 3.8 right?

mesutcelik:
👍  @noctarius , can you please resolve conflicts then we can merge it?

noctarius:
Hey, sorry was at the conference all day long.

PR #24, opened 2017-03-16
Improved dockerfile, health check scripts, documentation, bug fixes

lazerion:


hazelcast-openshift-rhel/healthcheck.sh
```diff
@@ -1,3 +1,14 @@
...
+HTTP_RESPONSE=$(curl --silent --write-out "HTTPSTATUS:%{http_code}" $URL)
+HTTP_BODY=$(echo $HTTP_RESPONSE | sed -e 's/HTTPSTATUS\:.*//g')
+HTTP_STATUS=$(echo $HTTP_RESPONSE | tr -d '\n' | sed -e 's/.*HTTPSTATUS://')
+
+if [ ! $HTTP_STATUS -eq 200  ]; then
+  echo "failure, http status: "$HTTP_STATUS
+  exit 1
+fi
+
+result=$(echo $HTTP_BODY | awk '{match($0,  /Hazelcast\:\:NodeState\=([A-Z]*)/, m)}END{print m[1]}')
```
noctarius: Nope we need the `clustersafe` state because otherwise there are still ongoing migration which would cause data loss when a node is restarted / shutdown too fast (like scale down or rolling update).

lazerion: I see, however when we add a new pod from OCP, `ClusterSafe` status becomes false, therefore all pods are restarted by Openshift itself, which also causes data loss, that was the best solution in my current understanding, do you have any solution alternatives? can we combine `NodeState` and `ClusterSafe` to get a liveness verdict?

noctarius: We might need to use both probs, clustersafe for readyness, nodestate for lifeness? does that make sense? :)

lazerion: Yes, agreed that we need to have different scripts for liveness and readiness.

hazelcast-openshift-rhel/hazelcast-template.js
```diff
@@ -35,7 +35,7 @@
 				},
 				"spec": {
 					"containers": [{
-						"image": "hazelcast/hazelcast-openshift-rhel",
+						"image": "hazelcast/openshift",
```
mesutcelik: why do we need to change this?

management-center/Dockerfile
```diff
@@ -1,11 +1,19 @@
 FROM java:7-jre
 ENV HZ_VERSION 3.8
 ENV HZ_HOME /opt/hazelcast/
+# Persistent VOLUME definitions
+ENV MANCENTER_DATA /data
+VOLUME ["/data"]
```
mesutcelik: We should mention this in a doc. `/data` needs to be available somewhere.

hazelcast-openshift-rhel/Dockerfile
```diff
@@ -9,14 +9,15 @@ ENV LANG en_US.utf8
...
+
 ARG HZ_KUBE_VERSION=1.0.0
 ARG HZ_MAVEN_DIR=${HZ_VERSION}
-ARG REPOSITORY_URL=https://repo1.maven.org/maven2
+ARG REPOSITORY_URL=https://repository-hazelcast-l337.forge.cloudbees.com
 
 LABEL name="hazelcast/hazelcast-openshift-rhel" \
       vendor="Hazelcast, Inc." \
-      version="7.2" \
+      version="${HZ_VERSION}" \
```
noctarius: As far as I remember this is supposed to be the REHL version used in the image

hazelcast-openshift-rhel/hazelcast-template.js
noctarius: good question

noctarius:
Shouldn't we configure Management Center in the hazelcast.xml? Question is, how do we discover the URL of Management Center?

management-center/Dockerfile
lazerion: will do

hazelcast-openshift-rhel/hazelcast-template.js
```diff
@@ -34,8 +35,16 @@
...
 				},
 				"spec": {
+					"volumes":[
+						{
+							"name":"hazelcast-storage",
+							"persistentVolumeClaim":{
+								"claimName":"hz-vc"
+							}
+						}
+					],
```
mesutcelik: I think this needs to go to documentation because first users will probably use the simplest approach. Adding volumes and customizing xml + custom jars would be more advanced usage.

The other option is to create another template with volume the volume option.

hazelcast-openshift-rhel/start.sh
```diff
@@ -23,12 +23,21 @@ if [ "$HAZELCAST_KUBERNETES_SERVICE_DNS" == "$HAZELCAST_KUBERNETES_SERVICE_NAME.
...
 echo "########################################"
 echo "# RUN_JAVA=$RUN_JAVA"
 echo "# JAVA_OPTS=$JAVA_OPTS"
-echo "# starting now...."
+echo "# CLASSPATH=$CLASSPATH"
 echo "########################################"
 
-java -server $JAVA_OPTS -Dhazelcast.http.healthcheck.enabled=true -Djava.net.preferIPv4Stack=true -Dhazelcast.enterprise.license.key=$HZ_LICENSE_KEY com.hazelcast.core.server.StartServer
+echo "Checking custom configuration"
+FILE=$HZ_DATA/hazelcast.xml
```
mesutcelik: user should now the name of the xml is `hazelcast.xml` before putting it there. This should be highlighted in the documentation. The other option is to accept the file name through the template.js but not sure if it is a good idea.

management-center/Dockerfile
noctarius: I think we should try to find a general naming set for this directories. In the new openshift docker file it is called _external_ (https://github.com/hazelcast/hazelcast-docker/blob/master/hazelcast-openshift-rhel/Dockerfile#L5) and is automatically added to the classpath.

hazelcast-openshift-rhel/start.sh
lazerion: Documentation in README.md will be updated accordingly.

hazelcast-openshift-rhel/hazelcast-template.js
lazerion: Documentation will be updated, users can input PersistentVolume that are **already created** by OCP itself.

hazelcast-openshift-rhel/Dockerfile
noctarius: Seems like I was wrong... can't find the reference I was looking for .... whatever :D

hazelcast-openshift-rhel/hazelcast-template.js
```diff
@@ -52,30 +61,39 @@
 						}, {
 							"name": "HAZELCAST_KUBERNETES_SERVICE_DOMAIN",
 							"value": "${KUBERNETES_SERVICE_DOMAIN}"
+						}, {
+							"name": "HZ_LICENSE_KEY",
+							"value": "${ENTERPRISE_LICENSE_KEY}"
+						},{
+							"name": "HZ_DATA",
+							"value": "/data/hazelcast"
```
noctarius: Still think we should be consistent between hazelcast and mancenter image

noctarius:
PS: I don't get the idea of this review thing, see comments, no idea if this is what it should be

hazelcast-openshift-rhel/hazelcast-template.js
lazerion: Mancenter image volume path will be consistent with hazelcast.

PR #48, opened 2017-12-22
changed base images to alpine

### LINKED ISSUE #36: Consider providing slim down version of Hazelcast images
Topic for discussion - follow up to #34.

Current `hazelcast/hazelcast` image size is `321MB`, but only `9MB` of it takes Hazelcast JAR itself. The image could be reduced by using a different base image. Then either only the reduced version would be provided by Hazelcast or both the full and the reduced.

The `openjdk` [Docker repository](https://hub.docker.com/_/openjdk/) offers `*-jre-slim` and/or `*-jre-alpine` images which could be used to reduce the size. The quick test shows following sizes of the final images:
* 285MB - hazelcast on the top of `openjdk:8-jre`
* 216MB - hazelcast on the top of `openjdk:8-jre-slim`
* 95MB - hazelcast on the top of `openjdk:8-jre-alpine` (with `bash` package added to the image)

mesutcelik:
fixes #36 

Smaller size of hazelcast images will be published as of 3.9.2 cc : @Serdaro


googlielmo:
Rebased on master

PR #64, opened 2018-05-12
Remove liveness.sh and readiness.sh from Kubernetes images

leszko:
Up from the version 3.10 the liveness.sh and readiness.sh are not needed, since the health endpoint can be called directly.


hazelcast-enterprise-kubernetes/README.md
```diff
@@ -161,9 +161,9 @@ spec:
         ports:
         - containerPort: 5701
         livenessProbe:
-          exec:
-            command:
-            - "./liveness.sh"
+          httpGet:
+            path: /hazelcast/health/node-state
+            port: 5701
```
mesutcelik: What do you think about putting a sample for readiness?

leszko: hmm... I'd wait until we decide if we recommend using readiness at all (as discussed in the OpenShift TDD).

Issue #62, opened 2018-05-11
Why have `server.sh` and `stop.sh` at all?


ps-jay:
Hi folks,

I'm wondering why there is `server.sh` and `stop.sh` in the containers?

Why not just run the `java -server ...` command in the foreground?

w0203j:
Yes I was wondering the same thing.

mesutcelik:
👍 

I think those can be removed with a proper documentation update. Before removing them, we need to move env variable definitions in the server.sh to Docker file and update documentation accordingly. Those need to be done in all Dockerfiles...

We would appreciate any contribution!

ps-jay:
Guess this is sort of related to #16, and maybe #42 as well.

I'll have a think on how to approach this.

ps-jay:
Closed via #66

Issue #42, opened 2017-11-14
Consider using graceful shutdown instead of terminate on SIGTERM


jerrinot:
Hazelcast registers a shutdown hook which is executed when JVM is about to shutdown. 
By default the hook is calling terminate, but it Docker environment it makes more sense to use a graceful shutdown. 

There a [group property](https://github.com/jerrinot/hazelcast/blob/b9cdf2c705a20a74639536661ebdd0db154078ca/hazelcast/src/main/java/com/hazelcast/spi/properties/GroupProperty.java#L353) for this. 

Also see https://docs.docker.com/engine/reference/commandline/stop/
Related to https://github.com/hazelcast/hazelcast-docker/issues/16

mesutcelik:
Apparently, we have to switch on `hazelcast.shutdownhook.policy=GRACEFUL` in every start script if we want to introduce graceful shutdown hook. However, I don't know consequences of this behavior like slowness in scale-in situations. 

`JAVA_OPTS` can be used in based image so need to change the behavior but not sure about k8s and openshift env. @googlielmo WDYT?

googlielmo:
I think we can keep the default behaviour, but I would add instructions in the documentation to show how to change that. I will open a PR for that if that's okay @jerrinot @mesutcelik

mesutcelik:
👍

ps-jay:
Hey @googlielmo ... Did you open that PR?  I can't find it.

ps-jay:
Now documented in README.md here: https://github.com/hazelcast/hazelcast-docker#stopping-a-hazelcast-member

Probably can close this issue.

leszko:
@jerrinot , are you fine to close the issue when it's documented in README how to gracefully shutdown?

jerrinot:
I am ok to have it just documented, but I'd like to see some guidance/considerations. 

Something along these lines: "TERMINATE means a container will be destroyed quickly, but it relies on backup for data safety. GRACEFUL shutdown will first trigger partition migration - which may take some time depending on your data size - and only then it quits. "  +perhaps a link Hazelcast Reference manual.

wdyt?

leszko:
👍 

Let's document it and then close this issue :)

Issue #16, opened 2017-01-18
removal of stop scripts from docker images


mesutcelik:
Instead of maintaining PID based stop script, we should rely on docker stop [docker stop](https://docs.docker.com/engine/reference/commandline/stop/)

see https://github.com/hazelcast/hazelcast-docker/pull/15

mesutcelik:
I think we will keep `stop.sh` in the repo in case someone wants to stop hazelcast via script before stopping Docker container. In normal conditions, we do not need stop.sh

ps-jay:
For anyone following this issue, the scripts have now been removed as part of #66.

PR #66, opened 2018-05-18
Remove `server.sh` and `stop.sh` scripts & rework Dockerfiles for OSS and Enterprise

ps-jay:
This pull request:
- removes the `server.sh` and `stop.sh` scripts
- reworks the `Dockerfile` for both the OSS and Enterprise images
- updates the `README.md` to match the changes

As discussed in #16 and #62, the `server.sh` and `stop.sh` scripts aren't really required.

Hazelcast itself can run in the foreground of the container, and catch & handle `SIGTERM` signals on its' own.

This is particularly true, when you consider the use of the shutdown hook functionality - as discussed in #42.

These changes are _mostly_ backward compatible, if you're already using these images.

The container still behaves as if it has the `server.sh` script, for example it still makes use of the environment variables `HZ_LICENSE_KEY`, `MIN_HEAP_SIZE`, etc.

The backward incompatible change is that `docker exec` to call `stop.sh` no longer works.  Hence why I've updated the `README.md` to also mention the shutdown hooks - as this might be important for some installations.

Summary of changes to the `Dockerfile` for OSS and Enterprise:
- Update Alpine, and install `bash` in a single `RUN` command to reduce final image size 
- Removing the apk cache after installing `bash` & `curl` to reduce the final image size
- Add `curl`, so that we can use `curl` to download Hazelcast, unzip it, and delete the zip file in a single `RUN` command, again reducing the final image size
- Do not extract the `code-samples`, `demo`, or `docs` into the image, these would generally not be needed for Hazelcast server runtime - also reducing final image size
- Use `curl` also for cache api jar download for consistency
- Move the logic from `server.sh` into the `CMD` command - it's not the most neat looking `CMD`, but it preserves backward compatibilty with images built with `server.sh` in them

When compared to the current publically available `hazelcast/hazelcast-enterprise:3.10` image (sha256:c7515234fcbe80b664bf7704973cee0c9d1b44bfc9623db44ebff3b2fe148a75) this new image final image size is 114MB, down from 421MB.


w0203j:
This is such a good improvements!

hasancelik:
@ps-jay LGTM 👍 Could you please sign our Contributor Agreement to have the PR merged into Hazelcast repository?

ps-jay:
Signed the CLA, and emailed it through.

leszko:
Thanks @ps-jay

Issue #53, opened 2018-01-15
cannot get endpoints in the namespace "default"


XieXianbin:
SEVERE: [172.20.215.2]:5701 [dev] [3.9.2] Failure executing: GET at: https://kubernetes.default.svc/api/v1/namespaces/default/endpoints/hazelcast. Message: Forbidden!Configured service account doesn't have access. Service account may have been revoked. endpoints "hazelcast" is forbidden: User "system:serviceaccount:default:default" cannot get endpoints in the namespace "default".
io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: GET at: https://kubernetes.default.svc/api/v1/namespaces/default/endpoints/hazelcast. Message: Forbidden!Configured service account doesn't have access. Service account may have been revoked. endpoints "hazelcast" is forbidden: User "system:serviceaccount:default:default" cannot get endpoints in the namespace "default".
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.requestFailure(OperationSupport.java:470)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.assertResponseCode(OperationSupport.java:407)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:379)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleResponse(OperationSupport.java:343)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleGet(OperationSupport.java:312)
	at io.fabric8.kubernetes.client.dsl.base.OperationSupport.handleGet(OperationSupport.java:295)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.handleGet(BaseOperation.java:787)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.getMandatory(BaseOperation.java:217)
	at io.fabric8.kubernetes.client.dsl.base.BaseOperation.get(BaseOperation.java:184)
	at com.hazelcast.kubernetes.ServiceEndpointResolver.resolve(ServiceEndpointResolver.java:81)
	at com.hazelcast.kubernetes.HazelcastKubernetesDiscoveryStrategy.discoverNodes(HazelcastKubernetesDiscoveryStrategy.java:102)
	at com.hazelcast.spi.discovery.impl.DefaultDiscoveryService.discoverNodes(DefaultDiscoveryService.java:74)
[...]

PR #71, opened 2018-05-23
RBAC authorization file

nordri:
Hello there,

I've struggled for a while with this issue and found there's a closed issue (#53) solving it, so I've basically created the rbac spec file (as the issue pointed to) and wrote a note on the README file to avoid future problems.

Cheers!


hasancelik:
Hi @nordri ,

Thanks a lot for your contribution! LGTM 👍 
Could you please sign our Contributor Agreement to have the PR merged into Hazelcast repository?

leszko:
Thanks @nordri for the contribution :)

I think it's a valuable information, but what do you think about putting it in the "Troubleshooting" section at the end of the README file? My feeling is that this is the step that may or may not be required for the your Kubernetes setup, so it should not be in the main flow of actions. Also, I'd rather see it as a description part, not necessarily the "hazelcast-rbac.yaml" file. Unless I'm wrong and you always need to set it up?

@googlielmo , What are your thoughts on that?

nordri:
Hi @hasancelik. you're welcome! I've just signed and sent the CLA doc ;)

@lazerion you're welcome! You may be right, but in my experience you always need to use RBAC authorization, so, IMHO, it should be a basic step in the deployment.

On the other hand, you mean a **description part** explaining what RBAC means and how it works?

Cheers!

googlielmo:
Hi @nordri 

Thanks for the contribution!

The wording in the PR itself seems correct to me (i.e., "you may need ...") as in general RBAC has to be enabled by the cluster admins.

Re the description part that @leszko mentioned, I believe he's suggesting to put the contents of `hazelcast-rbac.yaml` directly as an example configuration in the README itself, not necessarily adding it as a separate file.
Wdyt?

Cheers,
Guglielmo

leszko:
@nordri @googlielmo yeah, that's what I meant that it could be part of the description rather than a a file. Saying that, if you think you always need to configure it, then I'm fine to merge it as it is :)

nordri:
Well, I'm working with Kubernetes 1.9.3 and 1.10.0 and in both I had to use RBAC for different pourposes. Saying that, I'm pretty sure you will need to set up the authorization.

When you may not need to use RBAC? I can think, for example, an old Kubernetes installation or you installed it in advanced.

I've commited the new changes. LMK what you think!

Cheers!

nordri:
:smile:

PR #72, opened 2018-05-24
Add description about TERMINATE and GRACEFUL shutting down methods

### LINKED ISSUE #42: Consider using graceful shutdown instead of terminate on SIGTERM

leszko:
Changes:
- Add description to what `TERMINATE` and `GRACEFUL` means
- Remove `hazelcast.shutdownhook.enabled=true`, it's `true` by default anyway

close #42 

@jerrinot @ps-jay , please have a look.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,14 +111,16 @@ You can `stop` the member using the docker command: `docker stop <containerid>`.
 
 By default, Hazelcast is configured to `TERMINATE` on receiving the SIGTERM signal from Docker, which means that a container stops quickly, but the cluster's data safety relies on the backup stored by other Hazelcast members.
 
-The other option is to use the `GRACEFUL` shutdown, which triggers the partition migration before shutting down the Hazelcast member. Note that it may take some time depending on your data size. To use that approach, configure the following system property to your `JAVA_OPTS` environment variable:
+The other option is to use the `GRACEFUL` shutdown, which triggers the partition migration before shutting down the Hazelcast member. Note that it may take some time depending on your data size. To use that approach, configure the following properties:
 
-- `hazelcast.shutdownhook.policy=GRACEFUL`
-
-You may also want to configure:
-- `hazelcast.graceful.shutdown.max.wait=<seconds>`
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -109,12 +109,13 @@ Now you can run your own container with its ID or tag (if you provided `-t` opti
...
-By default Hazelcast is configured to `TERMINATE` on receiving the SIGTERM signal from Docker.
+By default, Hazelcast is configured to `TERMINATE` on receiving the SIGTERM signal from Docker, which means that a container stops quickly, but the cluster's data safety relies on the backup stored by other Hazelcast members.
 
-If you prefer a graceful exit, then add & configure these system properties to your `JAVA_OPTS` environment variable:
+The other option is to use the `GRACEFUL` shutdown, which triggers the partition migration before shutting down the Hazelcast member. Note that it may take some time depending on your data size. To use that approach, configure the following system property to your `JAVA_OPTS` environment variable:
 
-- `hazelcast.shutdownhook.enabled=true`
 - `hazelcast.shutdownhook.policy=GRACEFUL`
+
+You may also want to configure:
```
mesutcelik: I would mention `docker stop --time` here because docker timeout  must be equal or greater than `hazelcast.graceful.shutdown.max.wait`....

ps-jay: It's mentioned below on line 120, but it would be worth mentioning it can be configured up front in the `docker run` command by adding the `--stop-timeout` argument.

leszko: @mesutcelik @ps-jay I've reformatted the description. I think it's clearer now. Also added `docker run --stop-timeout`. PTAL.

Any comments are welcome, as always :)

ps-jay:
Nice one. 
Looks good to me : )

PR #79, opened 2018-06-14
Change RBAC from 'admin' to 'view'

leszko:
I don't think that 'admin' rights are required.


artem-zinnatullin:
This is much nicer, thanks @leszko 👍

PR #83, opened 2018-07-09
Add 'docker' Pardot ID and Expose to Dockerfiles

leszko:
Changes:
- Add Pardot ID Download Source `docker`
- Add `EXPOSE 5701`

This will help us trace how many people uses Hazelcast distributed as a Docker image.


hazelcast-oss/Dockerfile
```diff
@@ -47,8 +47,11 @@ RUN cd mvnw \
  && rm -rf ~/.m2 \
  && chmod -R +r $HZ_HOME
 
+# Set Pardot ID to 'docker'
+RUN echo 'hazelcastDownloadId=docker' > hazelcast-download.properties
+
 # Runtime constants
-ENV CLASSPATH_DEFAULT "${HZ_HOME}/*"
+ENV CLASSPATH_DEFAULT "${HZ_HOME}:${HZ_HOME}/*"
```
googlielmo: What's the rationale for having HZ_HOME itself on the classpath? (and why in this PR)

leszko: The added file `hazelcast-download.properties` is in ${HZ_HOME} and it's read from the classpath. I actually tried to add only this file to CP, but apparently it didn't work.

googlielmo: Got it, makes sense now! Thanks @leszko :-)

PR #98, opened 2018-12-07
Add template for Management Center in Hazelcast OSS

### LINKED ISSUE #85: hazelcast (3.7.8) not able to connect to mancenter
I am working on hazelcast version 3.7.8
Ran the docker image for mancenter 
`docker run -p 8080:8080 hazelcast/management-center:3.7.8`
Ran the docker image for hazelcast (specifying the mancenter URL) 
`docker run -e MANCENTER_URL="http://localhost:8080/mancenter" hazelcast/hazelcast:3.7.8`

hazelcast server is not able to connect mancenter

screen shot 2018-07-21 at 12 18 23 am

leszko:
fix #85

### DIFF
--- a/hazelcast-oss/Dockerfile
+++ b/hazelcast-oss/Dockerfile
@@ -47,16 +47,20 @@ RUN cd mvnw \
  && rm -rf ~/.m2 \
  && chmod -R +r $HZ_HOME
 
+ADD hazelcast.xml $HZ_HOME/hazelcast.xml
+
 # Set Pardot ID to 'docker'
 RUN echo 'hazelcastDownloadId=docker' > hazelcast-download.properties
 
 # Runtime constants
 ENV CLASSPATH_DEFAULT "${HZ_HOME}:${HZ_HOME}/*"
-ENV JAVA_OPTS_DEFAULT "-Djava.net.preferIPv4Stack=true"
+ENV JAVA_OPTS_DEFAULT "-Djava.net.preferIPv4Stack=true -Dhazelcast.mancenter.enabled=false"
 
 # Runtime environment variables
 ENV MIN_HEAP_SIZE ""
 ENV MAX_HEAP_SIZE ""
+ENV MANCENTER_URL ""
+
 ENV CLASSPATH ""
 ENV JAVA_OPTS ""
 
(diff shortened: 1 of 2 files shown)


hazelcast-oss/Dockerfile
```diff
@@ -69,6 +73,7 @@ CMD ["bash", "-c", "set -euo pipefail \
       && if [[ \"x${JAVA_OPTS}\" != \"x\" ]]; then export JAVA_OPTS=\"${JAVA_OPTS_DEFAULT} ${JAVA_OPTS}\"; else export JAVA_OPTS=\"${JAVA_OPTS_DEFAULT}\"; fi \
       && if [[ \"x${MIN_HEAP_SIZE}\" != \"x\" ]]; then export JAVA_OPTS=\"${JAVA_OPTS} -Xms${MIN_HEAP_SIZE}\"; fi \
       && if [[ \"x${MAX_HEAP_SIZE}\" != \"x\" ]]; then export JAVA_OPTS=\"${JAVA_OPTS} -Xmx${MAX_HEAP_SIZE}\"; fi \
+      && if [[ \"x${MANCENTER_URL}\" != \"x\" ]]; then export JAVA_OPTS=\"${JAVA_OPTS} -Dhazelcast.mancenter.enabled=true -Dhazelcast.mancenter.url=${MANCENTER_URL}\"; fi \
```
mesutcelik: It should normally be sufficient to enable MC just passing `-Dhazelcast.mancenter.url` as System Property so that we don't have to maintain hazelcast.xml...

It is actually not the only concern. The other one is if hazelcast.xml is provided by user, she sets MANCENTER_URL but forgets about adding those system properties into hazelcast.xml, then you end up with incorrect MC configuration. cc: @emre-aydin

emre-aydin: Should be easy to add those as group properties to members, but AFAIK introducing new group properties are generally frowned upon by the core team. WDYT @mmedenjak ?

mmedenjak: I don't like group properties but sometimes (rarely) I would allow it. On the other hand, you **DO NOT** want to mention adding group properties to Greg :)
Please refer to: https://hazelcast.atlassian.net/wiki/spaces/EN/pages/253953578/Hazelcast+Configuration

mesutcelik: I think we can keep what we have now but please keep in mind enabling MC is different than configuring hazelcast product. We might need to have an switch to on/off MC in cloud environments without touching hazelcast.xml. OTOH, if MC team implements auto discovery of hazelcast cluster, we will even not need any config on hazelcast side. It is best to wait for auto-discovery then. cc: @emre-aydin

PR #103, opened 2018-12-18
Add configuration for Travis CI (a simple PR builder)

kwart:
This pull request adds a simple Travis CI build configuration for both EE and OS Hazelcast docker images.

The main benefit is catching build problems in pull requests.


leszko:
I think the PR itself is fine, but I wonder if we want to support Travis CI? @Holmistr , what are the plans?
If we want to be Jenkins-centric, then I'd recommend changing it to Jenkinsfile.

Saying that, I think it's a good idea to have CI for Docker images. We will later even make it CD (when we work on the P/Z Linux PRDs).

mesutcelik:
Thanks for the PR @kwart !

The best part is actually having the configuration in github as opposed to JenkinsFile so it has a history and code review in the first place. 

I am ok to have mixed environments. I actually believe that teams can decide whatever CI they want to use and build the knowledge internally. 

I suggest to have a discussion with Jiri when he is back!

Holmistr:
Hi guys, I'm back! :)

I think that for such a simple thing, we can use Travis CI. However, I would rather have more complicated things in Jenkins. The reason is that we don't want to spend any time on supporting yet another CI environment. So once we encounter any problems with Travis, I will strongly vote for migrating this to Jenkins.

Holmistr:
If we already have place that we use for storing the Jenkinsfile, why don't we use it? We would keep the benefits of having the configuration in GH and not adding new environment to maintain. 

Didn't know about this.

kwart:
@leszko Can you share how the Jenkins integration works in the charts repository?
I don't see automatic builds used for PRs there.

For me the Travis CI behavior is more obvious. Nevertheless, for more complex or critical tasks is using a company preferred tool the safer way.

leszko:
Closing due to inactivity. Please reopen if needed.

kwart:
Reopening. I've spent some time trying a Jenkins equivalent yesterday, but I was not very successful.
Travis CI has several advantages:
* simpler configuration
* everyone knows what's going on
  * community can check the output of the job
  * job configuration in a single file within this repository
* in Travis CI, we don't experience Docker problems which we see in Jenkins
* if anything goes wrong with the Docker image build (e.g. too fat image) it won't affect our (Jenkins) build environment

Therefore I'm reopening this and asking for the reconsideration.

leszko:
I just spoke with Josef and he convinced me. The good points:
- Travis is integrated with GH, so we don't have to really maintain a new environment
- Travis output is visible for community, so there is no really a need for VPN, etc. and the community can read the results.

I actually now think we could move some of the OpenSource Hazelcast projects to travis, because sometimes the community sends a PR and don't know why it fails.

@Holmistr I added you as a reviewer, because after all the releases, PR builders are usually up to the QE team. @hasancelik you can also have a look.

mesutcelik:
Github Actions provide Travis like CI as far as I know. @pivovarit can give more insight.

pivovarit:
GH Actions are way faster and more reliable than Travis nowadays.
We use them here, for example: https://github.com/hazelcast-guides/hazelcast-quarkus/tree/master/.github/workflows

Holmistr:
Hi @leszko @kwart . Following point will hopefully no longer be valid after Dev Days, since it's will be my Project X:
> Travis output is visible for community, so there is no really a need for VPN, etc. and the community can read the results.

After consideration, I'm OK with using Travis CI for Docker images PR builder, since it provides very very simple capabilities. On the other hand, I'm strictly against following:
> I actually now think we could move some of the OpenSource Hazelcast projects to travis, because sometimes the community sends a PR and don't know why it fails

As I said, hopefully the point will not be valid anymore soon. I don't want to go the road that if it was a bit more difficult to setup in one environment, we will just create a new one. This leads to having several different environments etc., which we don't want to have.

I would allow Travis CI approach purely on Docker images, since the support is good, mature, easy to use. Also it's not a sophisticated test, it's basically a smoke test. And it's a bit problematic with our infrastructure to set this up in Jenkins (Docker in Docker problems etc.).

Approving under above mention conditions.

leszko:
@kwart will make a quick research on GH Actions and then we decide. I'm fine with both, but agree that the less environments we have, the better!

pivovarit:
From the security point of view, it doesn't require any extra 3rd party access

.travis.yml
```diff
@@ -0,0 +1,24 @@
+language: minimal
+
+services:
+  - docker
+
+env:
+  - HAZELCAST_IMAGE=hazelcast-enterprise
+  - HAZELCAST_IMAGE=hazelcast-oss
```
pivovarit: Doesn't this overwrite the entry above?

.travis.yml
```diff
@@ -0,0 +1,24 @@
...
+install:
+  - docker build -t $HAZELCAST_IMAGE $HAZELCAST_IMAGE
+
+script:
+  - docker images
+  - |
+    if [ "$HAZELCAST_IMAGE" = "hazelcast-oss" -o -n "$HZ_LICENSE_KEY" ]; then
+      docker run --name ${HAZELCAST_IMAGE}-container -e HZ_LICENSE_KEY=$HZ_LICENSE_KEY -d --rm $HAZELCAST_IMAGE
+      sleep 10
+      HZ_IP=$(docker inspect --format '{{ .NetworkSettings.IPAddress }}' ${HAZELCAST_IMAGE}-container)
```
pivovarit: Would be better to expose the port instead of digging for the container ip

.travis.yml
kwart: Nope, this starts 2 jobs - each with one of the `HAZELCAST_IMAGE` values.

pivovarit: That's very unintuitive and very cool at the same time :)

.travis.yml
kwart: Unfortunately not in combination with the subsequent simple NetCat check. The exposed port will be open even if there is nothing listening on the target port in container.

kwart:
We are out of luck with the GitHub Actions.
The workflow (in the 2nd commit) works on my fork but it doesn't work in hazelcast organization. @jerrinot Could you have a look?

image

pivovarit:
It's weird that the free plan is not even available

kwart:
Let's move with the Travis CI for now and if/once we upgrade the account let's switch to GH Actions then.

leszko:
Ok, merging. Thanks again @kwart . You did a really great job:
- Creating the Travis configuration
- Convincing everyone for this solution

And I don't remember any other PR having approves from 4 reviewers! 

Thanks again!

PR #107, opened 2019-03-11
Make user defined JAVA_OPTS take precedence over the default parameters

leszko:
Remove default `JAVA_OPTS` for the environment variable not passed.

There was an issue with Helm Charts which does not use `MANCENTER_URL`, but take care of Management Center configuration on their own.

### DIFF
--- a/hazelcast-enterprise/start-hazelcast.sh
+++ b/hazelcast-enterprise/start-hazelcast.sh
@@ -18,19 +18,21 @@ else
 fi
 
 if [ -n "${MIN_HEAP_SIZE}" ]; then
-  export JAVA_OPTS="${JAVA_OPTS} -Xms${MIN_HEAP_SIZE}"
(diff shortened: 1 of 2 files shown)


mesutcelik:
If you don't pass `hazelcast.mancenter.enabled`, what will be value here ? false by default?
https://github.com/hazelcast/hazelcast-docker/blob/master/hazelcast-enterprise/hazelcast.xml#L22

leszko:
> If you don't pass `hazelcast.mancenter.enabled`, what will be value here ? false by default?
> https://github.com/hazelcast/hazelcast-docker/blob/master/hazelcast-enterprise/hazelcast.xml#L22

You're actually right, we need to set it. Let me change it, so that the user defined values took the precedence over the deafult.

leszko:
Ok, changed. @mesutcelik , PTAL.

PR #109, opened 2019-03-22
Update pattern and help message

johnksv:
Fix help string: name is capture group 1 not 2.

Add colon to pattern to avoid.
Before:
`hazelcast_HazelcastInstance_clusterTime:_1553256269561{instance="hazelcast0c",name="_hzInstance_1_Test",type="HazelcastInstance"} 1553256269561` Occur on some beans. This make promethus treat each entry as its own time series (because of the underscore)
Now:
`hazelcast_HazelcastInstance_clusterTime:{instance="hazelcast0c",name="_hzInstance_1_Test",type="HazelcastInstance"} 1553256269561`

Update time series name
Before:
`hazelcast_mymapname_IMap_localOwnedEntryMemoryCost{instance="hazelcast0c",name="mymapname",type="IMap><>localOwnedEntryMemoryCost",} 0.0`
Now:
`hazelcast_IMap_localOwnedEntryMemoryCost{instance="hazelcast0c",name="mymapname",type="IMap><>localOwnedEntryMemoryCost",} 0.0`
I think this make more sense since you can use the `name` field to filter or aggregate.

Added underscore and dash to pattern, so beans with name e.g 'mytopic-event' also is matched


leszko:
@johnksv have you maybe already signed and send to us the Contributor's Agreement?

johnksv:
Thanks for the reminder. I had put it on hold since there was so much legal text. The CLA  is signed and sent.

PR #124, opened 2019-12-18
Improve Container Memory Limits support

leszko:
Changes:
- Set the default heap size as 80% of the total container memory size (before it was the default value 25%, which is way too low)
- Document usage of the container memory limits


leszko:
> 👍 Have you tried overriding this parameter for folks who want lower the percentage?

Yes, it works :)

PR #134, opened 2020-02-05
Fix 4.0 release

### LINKED ISSUE #116: Use YAML instead of XML for the default hazelcast configuration
We should do the following:
- Remove `hazelcast.xml` at all from `hazelcast-oss`. It's not needed anymore, since Management Center URL is not set for HZ > 4.0
- Change `hazelcast.xml` to `hazelcast.yaml` in `hazelcast-enterprise`

### LINKED ISSUE #127: Remove MANAGEMENT_CENTER URL > 4.0
Starting from Hazelcast `4.0`, there is no property `management-center` in the configuration. So we should:
- Remove `hazelcast.xml` from `hazelcast-oss`
- Remove `<management-center>` from `hazelcast-enterprise/hazelcast.xml`
- Remote MANAGEMENT_CENTER_URL from README

leszko:
Fix 4.0 release

Also: 
- Remove Management Center URL (fix #127)
- Change XML to YAML (fix #116)


mesutcelik:
I think we need to add how to use MC with Hazelcast 4. I did not see any such section in the README

leszko:
> I think we need to add how to use MC with Hazelcast 4. I did not see any such section in the README

I will add it as a separate PR. I need to sort it out and this PR is mainly to fix `4.0` and make the Docker image release happen.

mesutcelik:
What about 3.12.x users? There is no section explaining how to run MC and hazelcast. There might be some users still relying on 3.12.x so It would be useful to keep a section for them.

leszko:
> What about 3.12.x users? There is no section explaining how to run MC and hazelcast. There might be some users still relying on 3.12.x so It would be useful to keep a section for them.

Hmm, don't you think that the `3.12.z` branch is good enough? https://github.com/hazelcast/hazelcast-docker/tree/3.12.z

I think we should have the current info in the `master` branch.

mesutcelik:
We can at least give a reference to `3.12.z` branch in `master` branch readme.

leszko:
Added all mentioned instructions about Management Center here: #135

PR #137, opened 2020-02-26
WIP Allow to define logging format and class via env variables

galibey:
It is hard to override default logging output format without building new image.
In theory, it can be done by passing format string as system property to `JAVA_OPTS` variable,
but since the format often contains whitespaces and $ sign and `JAVA_OPTS` is passed through `eval` and at then end it is expanded in `exec` without double quotes, it is extremely hard to make it passing properly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,24 @@ $ docker run -e LOGGING_LEVEL=FINEST hazelcast/hazelcast
 
 Available logging levels are (from highest to lowest): `SEVERE`, `WARNING`, `INFO`, `CONFIG`, `FINE`, `FINER`, and `FINEST`. The default logging level is `INFO`.
 
+### LOGGING_FORMAT
+The logging output format can be changed using `LOGGING_FORMAT` variable. For example:
+
+```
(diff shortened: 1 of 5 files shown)


mesutcelik:
How about mounting a logging.properties externally instead of building a new image or creating such env variables?

Similar approach here for hazelcast config.
https://github.com/hazelcast/hazelcast-docker#using-custom-hazelcast-configuration-file

galibey:
@mesutcelik It is an option, right, but, it's inconvenient in some scenarios. For example, in our case, we have k18s pod which executes bash script that uses helm chart to run hazelcast-docker image on another k18s cluster. In this case it's easier to deal with env variables instead of creating files and pass them between clusters.

mesutcelik:
In this case, you can try to store logging.properties in a ConfigMap and attach to the pod. Don't get me wrong but I am just trying to figure out a generic solution beneficial to change every part of the file. 

If that does not work for you too then I am ok with this PR.

galibey:
Using logging.properties in a ConfigMap worked well. Thanks @mesutcelik for the advice.

mesutcelik:
Good to hear that. 🎉

PR #138, opened 2020-03-14
Corrected classpath environment variable

kurtbaker09:
The environment variable needs a * after it to get all the files that come from the CLASSPATH_EXT directory.


leszko:
Thanks @kurtbaker09

PR #142, opened 2020-05-05
Change `mvn dependency:copy` to static dependency list

leszko:
The result Docker image contains exactly the same files.

The motivation for this change is that we want to build Multi Arch Docker images and use [QEMU emulator](https://www.qemu.org/) for this. For some reason, at some architectures, `mvn dependency:copy` fails with Java 11 on some emulators.

### Pros & Cons

**Pros**:
- Build time is faster
- Executing `mvn dependency:copy` inside Dockerfile is hacky IMO

**Cons**:
- If you update `hazelcast-eureka` version, you need to manually update all its dependencies (we use `mvn dependency:copy` only for resolving Eureka plugin dependencies)

#### 1. Get rid of Eureka plugin
We may consider removing Eureka plugin from our Docker images. We actually forgot to update it for Hazelcast 4.0 and I didn't see anyone complaining about it. Or we can get rid of Eureka plugin until we rewrite it to have zero dependencies.

#### 2. Use multi stage build with Java 8 as builder
We can use [Multi-Stage Dockerfiles](https://docs.docker.com/develop/develop-images/multistage-build/) and then use Java 8 for `mvn dependency:copy`, but still use Java 11 for running Hazelcast.

This will most probably work, but running `mvn dependency:copy` on emulated environments is very slow, I was running it for 10 h on some AWS machine and it still wasn't completed.

#### 3. Do not use QEMU emulation
We could use real machines with different architectures, but currently nobody wants to provide them to us for free. So we have temporary testing machines, but not something we could use for automation.

#### 4. Remove Eureka plugin only in hazelcast-openshift
The real requirement is for `hazelcast-openshift`, so we could remove Eureka from `hazelcast-openshift` and keep it here.


hazelcast-enterprise/Dockerfile
```diff
@@ -3,15 +3,23 @@ FROM alpine:3.11
...
...1.1.jar https://repo1.maven.org/maven2/org/apache/httpcomponents/httpcore/4.4.6/httpcore-4.4.6.jar https://repo1.maven.org/maven2/com/fasterxml/jackson/core/jackson-core/2.9.4/jackson-core-2.9.4.jar https://repo1.maven.org/maven2/commons-codec/commons-codec/1.9/commons-codec-1.9.jar https://repo1.maven.org/maven2/com/google/code/findbugs/jsr305/3.0.1/jsr305-3.0.1.jar https://repo1.maven.org/maven2/com/netflix/netflix-commons/netflix-eventbus/0.3.0/netflix-eventbus-0.3.0.jar https://repo1.maven.org/maven2/commons-jxpath/commons-jxpath/1.3/commons-jxpath-1.3.jar https://repo1.maven.org/maven2/com/sun/jersey/contribs/jersey-apache-client4/1.19.1/jersey-apache-client4-1.19.1.jar https://repo1.maven.org/maven2/javax/servlet/servlet-api/2.5/servlet-api-2.5.jar https://repo1.maven.org/maven2/com/netflix/eureka/eureka-client/1.9.18/eureka-client-1.9.18.jar https://repo1.maven.org/maven2/com/netflix/archaius/archaius-core/0.7.5/archaius-core-0.7.5.jar https://repo1.maven.org/maven2/stax/stax-api/1.0.1/stax-api-1.0.1.jar https://repo1.maven.org/maven2/org/antlr/antlr-runtime/3.4/antlr-runtime-3.4.jar https://repo1.maven.org/maven2/com/sun/jersey/jersey-client/1.19.1/jersey-client-1.19.1.jar'
```
mesutcelik, line 22: How hard was to create that list?

leszko: Not really hard, you need to execute `mvn dependency:list -DincludeScope=runtime -DoutputAbsoluteArtifactFilename=true`. Plus change the local prefix path to the remote URL.

leszko:
> Approving to allow resolving the issue.
> Still, from my PoV is the `dependency:copy` a less hacky way than the direct download. It's not just about the Eureka plugin, but also Netty tcnative libs. With Maven involved, a library version update resolves everything. Without Maven, we should check also the changes in the library's POM and POMs of all transitive dependencies.

I agree @kwart . You're right, in terms of the maintenance, the `dependency:copy` is better. I think that if we decide that resolving these dependencies manually is really a problem, then I'll just include here a script to fetch dependencies and the other step would be to use some Python tool to fetch maven dependencies. I think that some tools like these exist, but are not really mature.

leszko:
> I believe what you implemented is right way to do. We are not frequently changing Eureka version so it is good to go for now.
> Can you please a github issue for dependency free Eureka plugin?
> 
> I also openshift is not even using eureka so your new issue can mention updating Openshift with Eureka when we remove dependencies.

I checked and the OpenShift image does not really include Eureka.

Created an issue in `hazelcast/hazelcast-eureka` repository. We can work on that when we have a spare moment. https://github.com/hazelcast/hazelcast-eureka/issues/50

PR #143, opened 2020-05-08
Add Jenkinsfile for the release automation

leszko:
Add Docker image release automation.

Before the images were built by the Docker Hub automation, but it does not support multi-arch builds, so we need to automate it ourselves.

- I've already turned-off [Docker Hub Automation](https://hub.docker.com/repository/docker/hazelcast/hazelcast/builds) (so the images won't be built after every tag pushed to the repository)
- @lazerion @hasancelik could I ask you to add a trigger for the [hazelcast-docker-release pipeline](http://jenkins.hazelcast.com/view/Plugin%20Release/job/hazelcast-docker-release/)? It should be triggered after you released a new version of Hazelcast and after you pushed a new tag to [hazelcast/hazelcast-docker](url) repository. The pipeline has one parameter `HAZELCAST_DOCKER_TAG` and it should be the Hazelcast version (like the tag name, but without `v` at the beginning), for example: `4.0.1`
- I'll add a similar pipeline for `hazelcast-openshift`, but we need first to solve the issue with publishing multi-arch images to Red Hat Registry, so no changes there for now
- I'll extend this `Jenkinsfile` with tests and failure mail notifications in a separate PR. For now, I'd like to have exactly what we had with the use of Docker Hub automation

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -22,7 +22,7 @@ pipeline {
             steps {
                 dir("./hazelcast-oss") {
                     script {
-                        sh "docker buildx build -t leszko/hazelcast:${HAZELCAST_DOCKER_TAG} --platform=linux/arm64,linux/amd64,linux/ppc64le,linux/s390x . --push"
+                        sh "docker buildx build -t hazelcast/hazelcast:${HAZELCAST_DOCKER_TAG} --platform=linux/arm64,linux/amd64,linux/ppc64le,linux/s390x . --push"
                     }
 
                 }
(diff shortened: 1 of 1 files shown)


hasancelik:
> could I ask you to add a trigger for the hazelcast-docker-release pipeline? It should be triggered after you released a new version of Hazelcast and after you pushed a new tag to hazelcast/hazelcast-docker repository. The pipeline has one parameter HAZELCAST_DOCKER_TAG and it should be the Hazelcast version (like the tag name, but without v at the beginning), for example: 4.0.1

@leszko  Afaik, triggering jenkins job with some parameters is required some additional jenkins cli or something like that which should be utilized by our release scripts/pipeline. For fully automated process,  Hazelcast version should be retrieved dynamically from tag's Dockerfile itself and image should be built based on this version. 

Here is the correct configuration for tag-based trigger mechanism btw:
https://stackoverflow.com/questions/29742847/jenkins-trigger-build-if-new-tag-is-released
Screen Shot 2020-05-11 at 11 19 38

lazerion:
@leszko 
+1 from my end for @hasancelik comment. As we currently have `Docker File Updates` as **last** step of release pipeline, I think it is proper atm to use that approach.

Please note that regarding last step executed only pipeline successfully completed and nothing to rollback.

leszko:
Closing this one. Jenkinsfile does not work with automatic builds upon a new tag push.

So, I'll just create a Freestyle job which will be triggered automatically. So no change is required on your side, @lazerion.

The pipeline is available here: http://jenkins.hazelcast.com/view/Plugin%20Release/job/hazelcast-docker-release/

PR #144, opened 2020-05-14
Add mail notifications about Travis build failure

leszko:


.travis.yml
```diff
@@ -22,3 +22,12 @@ script:
       # license key is not available in PRs (to avoid leaking the secret value), lets just check we really run the EE
       docker run --name ${HAZELCAST_IMAGE}-container --rm $HAZELCAST_IMAGE 2>&1 | grep -q "Hazelcast Enterprise"
     fi
+ 
+notifications:
+  email:
+    recipients:
+      - rafal@hazelcast.com
+      - mesut@hazelcast.com
+      - hasan@hazelcast.com
```
kwart, line 31: What about using a single mail alias here? So changes in Hazelcast teams don't require Travis config updates.

leszko: yeah, good idea, changed to `integration@hazelcast.com`.

kwart: Thx.

leszko: I had to revert back to the direct mail addresses, since you cannot use address that is not registered in GH.

https://docs.travis-ci.com/user/notifications/#missing-build-notifications

PR #149, opened 2020-06-10
Remove version from jmx_prometheus_javaagent file name (#148)

### LINKED ISSUE #148: Enabling prometheus metrics stops starting the image
adding -e PROMETHEUS_PORT=8080 is causing the image not to start as the start script refers to the jar without version number added. The Dockerfile run commands puts it with a version number in the file name to $HZ_HOME/lib.

```
ls $HZ_HOME/lib/jmx_prometheus_*
/opt/hazelcast/lib/jmx_prometheus_javaagent-0.11.0.jar
```

start-hazelcast.sh
```
...
export JAVA_OPTS="-javaagent:${HZ_HOME}/lib/jmx_prometheus_javaagent.jar=${PROMETHEUS_PORT}
...
```

Dockerfile
```
...
for JAR_URL in ${HAZELCAST_ALL_URL} ${CACHE_API_URL} ${PROMETHEUS_AGENT_URL} ${EUREKA_PLUGIN_URLS}; do curl -sf -O -L ${JAR_URL}; done
...
```

maurizio-lattuada:
Fixed #148 by removing the version value from file name

### DIFF
--- a/hazelcast-enterprise/Dockerfile
+++ b/hazelcast-enterprise/Dockerfile
@@ -49,6 +49,7 @@ RUN echo "Updating Alpine system" \
     && mkdir "${HZ_HOME}/lib" \
     && cd "${HZ_HOME}/lib" \
     && for JAR_URL in ${HAZELCAST_ALL_URL} ${CACHE_API_URL} ${PROMETHEUS_AGENT_URL} ${NETTY_URL} ${NETTY_TCNATIVE_URL} ${EUREKA_PLUGIN_URLS}; do curl -sf -O -L ${JAR_URL}; done \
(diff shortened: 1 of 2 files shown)


maurizio-lattuada:
Sorry for the dumb question (especially if this is not the right place to ask it): is there an ETA when this new Hazelcast Docker image will be available on Docker Hub?
Just asking because I do not really know how the whole process works.

Thank you

hasancelik:
Hi @maurizio-lattuada,

It is fine totally fine :) `latest` tag is already updated with your changes:
https://hub.docker.com/repository/docker/hazelcast/hazelcast/tags?page=1&name=latest

If you are asking about new stable image like `4.0.2`, I can not give exact date but probably it will be available in next few weeks.

Thanks for the PR, btw.

PR #153, opened 2020-08-03
Add vulnerability scanning

leszko:
Scanning Docker images by vulnerability scanners.

Sample output: https://github.com/leszko/hazelcast-docker/runs/939978581

Note that they just print the information and won't prevent the images from the building.

We need 2 separate scanners, because:
- Azure detects better for some Code Style issues
- Anchore detects the jars with vulnerabilities

### DIFF
--- a/.github/workflows/latest_image_push.yml
+++ b/.github/workflows/latest_image_push.yml
@@ -31,3 +31,20 @@ jobs:
           docker buildx build --push \
             --tag hazelcast/hazelcast-enterprise:latest \
             --platform=linux/arm64,linux/amd64,linux/ppc64le,linux/s390x hazelcast-enterprise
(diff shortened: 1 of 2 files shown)


kwart:
@leszko Could you create also a scheduled Github action with the checks (executed weekly or daily) and alerting included (e.g. email)?
I'm afraid the issues can be missed if they are logged only.

leszko:
@kwart It's a good idea. But how about adding it later after we fix all vulnerabilities? I don't want to flood us with failing builds right now.

So the plan is:
- Merge this one
- Fix vulnerabilities
- Set up a daily build

Wdyt?

kwart:
Sure, we can go in more steps :+1:

PR #154, opened 2020-08-03
Fix Eureka dependencies vulnerabilities

leszko:


kwart:
The `curl` calls should be improved to fail the Docker build if they fail to download any of the JARs.
Maybe something like:
```bash
curl -sf -O -L ${JAR_URL} || exit $?
```

leszko:
> ```shell
> curl -sf -O -L ${JAR_URL} || exit $?
> ```

Good idea. Updated, PTAL. @kwart

PR #162, opened 2020-09-17
Include log4j2.

### LINKED ISSUE #129: Consider including logging libraries  (eg. log4j, log4j2, logback)
Related to the issue created in the Helm Chart repository: https://github.com/hazelcast/charts/issues/53

enozcan:
Closes #129


hazelcast-enterprise/Dockerfile
```diff
@@ -46,10 +48,11 @@ RUN echo "Installing new APK packages" \
     && echo "Downloading Hazelcast and related JARs" \
     && mkdir "${HZ_HOME}/lib" \
     && cd "${HZ_HOME}/lib" \
-    && for JAR_URL in ${HAZELCAST_ALL_URL} ${CACHE_API_URL} ${PROMETHEUS_AGENT_URL} ${NETTY_URL} ${NETTY_TCNATIVE_URL} ${EUREKA_PLUGIN_URLS}; do curl -sf -O -L ${JAR_URL} || exit $?; done \
+    && for JAR_URL in ${HAZELCAST_ALL_URL} ${CACHE_API_URL} ${PROMETHEUS_AGENT_URL} ${NETTY_URL} ${NETTY_TCNATIVE_URL} ${EUREKA_PLUGIN_URLS} ${LOG4J2_URLS}; do curl -sf -O -L ${JAR_URL} || exit $?; done \
     && mv jmx_prometheus_javaagent-*.jar jmx_prometheus_javaagent.jar \
-    && echo "Granting read permission to ${HZ_HOME}" \
+    && echo "Granting read and write permissions to ${HZ_HOME}" \
     && chmod -R +r ${HZ_HOME} \
+    && chmod o+w ${HZ_HOME} \
```
hasancelik, line 55: Why `write` permission is required?

enozcan: Otherwise `sed` fails during in-place editing at:
```
if [ -n "${LOGGING_LEVEL}" ]; then
  sed -i "s/rootLogger.level=info/rootLogger.level=${LOGGING_LEVEL}/g" log4j2.properties
fi
```
The currently documented command also faces this problem when it's not run as root.

hasancelik (approved):
👍

enozcan:
Thanks for the review @hasancelik.
Do you have any idea why the builder does not fail on other architectures but s390x? TBH the logs do not make sense to me: https://travis-ci.org/github/hazelcast/hazelcast-docker/jobs/728021165

hazelcast-enterprise/start-hazelcast.sh
```diff
@@ -22,8 +22,7 @@ if [ -n "${PROMETHEUS_PORT}" ]; then
 fi
 
 if [ -n "${LOGGING_LEVEL}" ]; then
-  sed -i "s/java.util.logging.ConsoleHandler.level = INFO/java.util.logging.ConsoleHandler.level = ${LOGGING_LEVEL}/g" logging.properties
-  sed -i "s/.level= INFO/.level= ${LOGGING_LEVEL}/g" logging.properties
+  sed -i "s/rootLogger.level=info/rootLogger.level=${LOGGING_LEVEL}/g" log4j2.properties
```
leszko, line 25: Consider using variable substitution (instead of sed) as described here: https://dougbreaux.github.io/2013/06/05/log4j-log-level.html

hazelcast-enterprise/log4j2.properties
```diff
@@ -0,0 +1,15 @@
...
+
+appender.console.type=Console
+appender.console.name=STDOUT
+appender.console.layout.type=PatternLayout
+appender.console.layout.pattern=%d [%highlight{${LOG_LEVEL_PATTERN:-%5p}}{FATAL=red, ERROR=red, WARN=yellow, INFO=green, DEBUG=magenta}] [%style{%t{1.}}{cyan}] [%style{%c{1.}}{blue}]: %m%n
+
+rootLogger.level=info
+rootLogger.appenderRef.stdout.ref=STDOUT
+
+# logger.internal.name=com.hazelcast.internal
```
leszko, line 11: I wonder if we need this commented-out lines. What are they for?

leszko (commented):
Goog job Enes! I really like the new logs!

Added 2 comments.

hazelcast-enterprise/start-hazelcast.sh
enozcan: Done

hazelcast-enterprise/log4j2.properties
enozcan: They were examples of custom configurations like we currently have:
https://github.com/hazelcast/hazelcast-docker/blob/4be9a22304186153a8f4fc55b1ef661875c47bd4/hazelcast-oss/logging.properties#L51-L58
However, these are log4j2 related properties, and including only a few examples does not make much sense as there are many others. Hence I removed completely.

hazelcast-enterprise/Dockerfile
enozcan: Removed with environment variable sol'n.

leszko (approved):

enozcan:
s390x arch forms a super weird behavior for logging pattern. For instance, I couldn't make it work to print logger names with abbreviations (i.e. `[c.h.c.CPSubsystem]`). It needs `%c{1.}` pattern that causes the container to crash on s390x with SIGFPE. Moreover, where `[DEBUG][main]` is fine, having a space between them `[DEBUG] [main]` causes it to crash again. I haven't seen any reported issue similar to this but come up with these by trial and error.  🤯

So I have 2 suggestions:

- Having a common pattern for all archs compatible with s390x.

  - **pros:** no extra arch dependent work is needed
  - **cons:** logs will lose abbreviation format for logger names. (i.e.`[c.h.c.CPSubsystem]` becomes `[com.hazelcast.cp.CPSubsystem]` vice versa). 

- Changing format via env vars (as we do for logging level) when s390x is detected.

  - **pros:** other architectures will not affected by this.
  - **cons:** some ugly works needed in start.sh

@leszko Do you have any further suggestions? WDYT?

leszko:
> s390x arch forms a super weird behavior for logging pattern. For instance, I couldn't make it work to print logger names with abbreviations (i.e. `[c.h.c.CPSubsystem]`). It needs `%c{1.}` pattern that causes the container to crash on s390x with SIGFPE. Moreover, where `[DEBUG][main]` is fine, having a space between them `[DEBUG] [main]` causes it to crash again. I haven't seen any reported issue similar to this but come up with these by trial and error. 🤯
> 
> So I have 2 suggestions:
> 
> * Having a common pattern for all archs compatible with s390x.
>   
>   * **pros:** no extra arch dependent work is needed
>   * **cons:** logs will lose abbreviation format for logger names. (i.e.`[c.h.c.CPSubsystem]` becomes `[com.hazelcast.cp.CPSubsystem]` vice versa).
> * Changing format via env vars (as we do for logging level) when s390x is detected.
>   
>   * **pros:** other architectures will not be affected by this.
>   * **cons:** some ugly works needed in start.sh
> 
> @leszko Do you have any further suggestions? WDYT?

That's all very weird. I'd go with the second option. Because we don't know of anyone even using Hazelcast on s390x, so I'd not adjust anything for that architecture. If it's possible to detect it and use different logging for s390x docker image, then I'd go for that. However, I'm not sure what this `ugly works` is. Do you already know how this ugly IF will look like?

enozcan:
Not exactly but something like:

```bash
BASE="%d [%highlight{${LOG_LEVEL_PATTERN:-%5p}}{FATAL=red, ERROR=red, WARN=yellow, INFO=green, DEBUG=magenta}]"
ARCH=`arch`

if [[ "$ARCH" == "s390x" ]]; then
  export LOGGING_FORMAT="$BASE [%style{%t{1.}}{cyan}] [%style{%c{1.}}{blue}]: %m%n"
else
  export LOGGING_FORMAT="$BASE[%style{%t{1.}}{cyan}] [%style{%-10c}{blue}]: %m%n"
fi
```

leszko:
> Not exactly but something like:
> 
> ```shell
> BASE="%d [%highlight{${LOG_LEVEL_PATTERN:-%5p}}{FATAL=red, ERROR=red, WARN=yellow, INFO=green, DEBUG=magenta}]"
> ARCH=`arch`
> 
> if [[ "$ARCH" == "s390x" ]]; then
>   export LOGGING_FORMAT="$BASE [%style{%t{1.}}{cyan}] [%style{%c{1.}}{blue}]: %m%n"
> else
>   export LOGGING_FORMAT="$BASE[%style{%t{1.}}{cyan}] [%style{%-10c}{blue}]: %m%n"
> fi
> ```

I think this is ok.

enozcan:
@leszko The build is now passing. PTAL.

leszko:
> @leszko The build is now passing. PTAL.

Looks good from my end 👍

hasancelik (approved):
👍

PR #164, opened 2020-09-18
Add scheduled vulnerability check.

### LINKED ISSUE #160: [Continuous Integration] Add daily job to check for vulnerabilities
We already have a [GH Action](https://github.com/hazelcast/hazelcast-docker/blob/master/.github/workflows/latest_image_push.yml) which checks for vulnerabilities after each push to master. It prints all vulnerabilities found in Hazelcast Docker image.

Now, we need to create similar GH Actions which:
- runs every day
- checks for vulnerabilities
- fails when a vulnerability is found
- notifies us by mail

enozcan:
Closes #160

### DIFF
--- a/.github/workflows/nightly_vulnerability_scan.yml
+++ b/.github/workflows/nightly_vulnerability_scan.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Build EE image
         run: |
           docker build -t hazelcast/ee:${{ github.sha }} hazelcast-enterprise
-          
+
       - name: Scan OSS image by Azure (Trivy + Dockle)
         uses: Azure/container-scan@v0
         with:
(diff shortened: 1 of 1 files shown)


hasancelik (approved):

leszko (commented):
Thanks Enes for the PR!

Two questions:
1. Does the build fail in case a vulnerability is found?
2. Will it send a mail to us? :)

enozcan:
For `anchore/scan-action` I added the required parameter (per [doc](https://github.com/anchore/scan-action#failing-the-build-on-policy-andor-vulnerabilities)) to fail. Also I have tested `Azure/container-scan` with `alpine:3.9.2` and seemingly it fails without any configuration needed (related [job](https://github.com/enozcan/hazelcast-docker/runs/1135988229)).

About the mailing issue, I'm not 100% sure who exactly receives mail notifications on failure for the actions in an organization's repo. This might be related to one's notification settings (may be watchers as well). But I found a recent issue here:

https://github.community/t/how-to-control-notifications-for-cron-triggered-workflows/117652

and I'm not sure if we want to add any of these third party actions - which are not that popular (~50 stars at most), to send mails.

leszko (approved):

PR #167, opened 2020-10-01
Fix notification email.

enozcan:


hasancelik (approved):
I was very happy without notification emails 😞 anyway thanks @enozcan 😄

PR #168, opened 2020-10-01
Add arminc/clair-scanner to GH actions.

enozcan:
Adds an extra checker to the current pipeline for possible vulnerabilities.

Screen Shot 2020-10-01 at 17 13 40


leszko (approved):
Wow, it's super complicated to run the clair scan... but I'm fine with that. If we see that we have some issues with the scanner itself (because we use `latest` images), then we can re-consider having it.

But for now, LGTM from my side 👍 

Let's wait what @hasancelik thinks :)

enozcan:
I agree. For the db image, we need to keep it latest as it is updated daily. But the other images are latest dependent and not necessarily stable.

hasancelik (approved):
Great job @enozcan 👍 I think we should try Clair for a while as Rafal said. I hope it will work without any issue because Clair is the best fir our docker images. Tanzu PKS and RedHat Connect Registry uses this tool as default scanning mechanism.

enozcan:
Guys, after the talk I reconsidered the PR and decided not to merge this for now. I'm planning to build an action with stable versions and a few more improvements. And then make a release from which not only we but community can also benefit.
So, let me close this now and continue with #170. Thanks for your reviews anyway as I'm sorry about that.

PR #126, opened 2020-01-03
Upgrade to Java 11

leszko:
Update to Java 11 and Alpine 3.11.

Comparison of image sizes

|                    | Java 8 (current) | Java 11 (this PR) |
|--------------|------------------|-------------------|
| hazelcast    |   78.97 MB         | 91.33 MB                        |
| hazelcast-enterprise |  85.28 MB  | 97.85 MB |

**Pros** of moving to Java 11:
- Keeping image up to date
- Using new features from Java 11 (like better GC, faster TLS)

**Cons** of moving to Java 11:
- (slightly) bigger docker images
- Non-backwards-compatibility in terms of Java internals
- (Alpine Linux version related) Netty-tcnative is compiled against the old (1.0.0) version of `libssl` which is not included in Alpine 3.9 and newer. We would need to use a version from Alpine 3.8.

------------------------------------------------------

[**archived below**] Issues already solved thanks to @kwart

I try to upgrade Alpine version and Java version, but I have a problem with OpenSSL. It fails with the exception below. The issue is with Alpine upgrade. It's fails even if I just upgraded Alpine to `11` and kept Java 8.

@kwart you're an expert here, could you help?

```
Exception in thread "main" java.lang.UnsatisfiedLinkError: failed to load the required native library
        at io.netty.handler.ssl.OpenSsl.ensureAvailability(OpenSsl.java:474)
        at com.hazelcast.nio.ssl.OpenSSLEngineFactory.init(OpenSSLEngineFactory.java:76)
        at com.hazelcast.nio.ssl.AbstractTLSChannelInitializer.loadSSLEngineFactory(AbstractTLSChannelInitializer.java:47)
        at com.hazelcast.nio.ssl.AbstractTLSChannelInitializer.<init>(AbstractTLSChannelInitializer.java:25)
        at com.hazelcast.nio.ssl.UnifiedTLSChannelInitializer.<init>(UnifiedTLSChannelInitializer.java:51)
        at com.hazelcast.nio.EnterpriseChannelInitializerProvider.createUnifiedTlsChannelInitializer(EnterpriseChannelInitializerProvider.java:94)
        at com.hazelcast.nio.EnterpriseChannelInitializerProvider.<init>(EnterpriseChannelInitializerProvider.java:37)
        at com.hazelcast.instance.EnterpriseNodeExtension.createChannelInitializerProvider(EnterpriseNodeExtension.java:615)
[...]

PR #171, opened 2020-10-05
Upgrade openssl version.

enozcan:
Upgrades openssl version to 1.1.1g-r0 as suggested by the vulnerability [check](https://github.com/hazelcast/hazelcast-docker/runs/1209332122). (default version at alpine:3.11 pkg [repo](https://pkgs.alpinelinux.org/package/v3.11/main/x86_64/openssl))

I see the discussions made here at #126. But I didn't see any problem with the given example in README:
```
docker run -v `pwd`:/keystore -e HZ_LICENSE_KEY=<your_license_key> \
    -e TLS_ENABLED=true \
    -e "JAVA_OPTS=-Djavax.net.ssl.keyStore=/keystore/server.keystore -Djavax.net.ssl.keyStorePassword=123456 -Djavax.net.ssl.trustStore=/keystore/server.truststore -Djavax.net.ssl.trustStorePassword=123456" \
    hazelcast/hazelcast-enterprise
```

as members form a cluster succesfully. Is there another way to reproduce the case you reported in the PR @leszko?

### DIFF
--- a/hazelcast-enterprise/Dockerfile
+++ b/hazelcast-enterprise/Dockerfile
@@ -42,9 +42,7 @@ COPY *.sh *.yaml *.jar *.properties ${HZ_HOME}/
 
 # Install
 RUN echo "Installing new APK packages" \
-    && apk add --no-cache openjdk11-jre bash apr curl procps nss \
-    && echo "Installing old OpenSSL 1.0 from Alpine 3.8" \
-    && apk add --no-cache --repository http://dl-cdn.alpinelinux.org/alpine/v3.8/main/ openssl=1.0.2u-r0 \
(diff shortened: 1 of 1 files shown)


hasancelik:
@enozcan I reproduced the issue during TKGI PRD with steps at [the issue that created by](https://github.com/netty/netty-tcnative/issues/518) @kwart. I think related issue still exists.

enozcan:
Thanks @hasancelik. I see the latest tcnative (v2.0.34) is compiled against libssl 1.0 as well. I think we need to ignore these vulnerabilities during scan for a while.

enozcan:
Closing hence the problem still exists with the latest tcnative (v2.0.34).

PR #173, opened 2020-10-07
Ignore vulnerabilities from libssl 1.0

enozcan:
**- Ignores libssl vulnerabilities for Snyk check for 3 months.** 
After this expires, we can try to upgrade openssl version if the latest is compatible or else extend the time period. See #171, #126.

**- Fixes PR builder** 
Uses `on: pull_request_target` to fix authentication problem caused by the missing secrets on fork builds.

~~**- Removes PR builder** 
Snyk authentication fails for PRs on `on: pull_request` action due to the missing token hence has no effect but blocks merging. The secret might be visible to fork PRs when `on: pull_request_target` is used instead (see [doc](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request_target)).~~

### DIFF
--- a/.github/containerscan/.snyk
+++ b/.github/containerscan/.snyk
@@ -1,0 +1,43 @@
+version: v1.5.0
+ignore:
+  SNYK-ALPINE311-OPENSSL-544606:
+    - '*':
+        reason: "netty-tcnative requires libssl 1.0"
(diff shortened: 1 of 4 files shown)


leszko (approved):

hasancelik (approved):
👍

PR #177, opened 2020-10-27
Unification of DockerHub and Openshift EE image

hasancelik:
Maintining two different repos/Dockerfiles for almost identical EE images does not make sense. Only disadvantage is increase of image size:

Screen Shot 2020-12-18 at 12 20 28

Related PRD --> https://hazelcast.atlassian.net/wiki/spaces/PM/pages/2677604456/Unification+of+RedHat+Connect+and+DockerHub+IMDG+Enterprise+Images

At the end of the process, we can remove [hazelcast/hazelcast-openshift](https://github.com/hazelcast/hazelcast-openshift) repo, create images for RedHat registry from [hazelcast/hazelcast-docker](https://github.com/hazelcast/hazelcast-docker)

I have created [temp image](https://hub.docker.com/layers/hasancelik/hazelcast-enterprise/ubi8-minimal/images/sha256-37ca72a97d8a4e13abafeff6b282073c8b33de875d505dd5a5cdee0ce9fd30e8?context=explore) for test purposes.

fixes https://github.com/hazelcast/hazelcast-openshift/issues/39


.travis.yml
```diff
@@ -16,6 +16,10 @@ env:
   - HAZELCAST_IMAGE=hazelcast-oss
 
 install:
+  - |
+    if [ "$HAZELCAST_IMAGE" = "hazelcast-enterprise"]; then
+      docker login registry.redhat.io -u ${RH_REGISTRY_USERNAME} -p ${RH_REGISTRY_PASSWORD}
```
leszko, line 21: One drawback of this unification would be that you need to have access to Red Hat Registry if you want to build the image.

hazelcast-enterprise/Dockerfile
```diff
@@ -1,11 +1,11 @@
-FROM alpine:3.11
+FROM registry.redhat.io/ubi8/ubi
 
 # Versions of Hazelcast and Hazelcast plugins
 ARG HZ_VERSION=4.0.3
 ARG CACHE_API_VERSION=1.1.1
 ARG JMX_PROMETHEUS_AGENT_VERSION=0.11.0
-ARG NETTY_VERSION=4.1.44.Final
-ARG NETTY_TCNATIVE_VERSION=2.0.28.Final
+ARG NETTY_VERSION=4.1.47.Final
```
leszko, line 7: Have you tested if OpenSSL works correctly with the updated versions?

hazelcast-enterprise/Dockerfile
```diff
@@ -35,32 +39,44 @@ ENV HZ_HOME="${HZ_HOME}" \
     CLASSPATH="" \
     JAVA_OPTS=""
 
+LABEL name="Hazelcast IMDG Enterprise" \
+      maintainer="info@hazelcast.com" \
+      vendor="Hazelcast, Inc." \
+      version="8.1" \
```
leszko, line 45: What is this `version`? I know we had it in `hazelcast-openshift`, but is it still mandatory to have?

hazelcast-enterprise/Dockerfile
```diff
@@ -35,32 +39,44 @@ ENV HZ_HOME="${HZ_HOME}" \
...
+      release="${HZ_VERSION}" \
+      url="http://www.hazelcast.com" \
+      summary="Hazelcast IMDG Enterprise Image, certified to RHEL 8.1" \
```
leszko, line 48:
```suggestion
      summary="Hazelcast IMDG Enterprise Image, certified to RHEL" \
```

To reduce the maintainance burden.

hazelcast-enterprise/Dockerfile
```diff
@@ -35,32 +39,44 @@ ENV HZ_HOME="${HZ_HOME}" \
...
+      description="Starts a standalone Hazelcast server instance to form a cluster based on Kubernetes discovery inside Openshift/Kubernetes"
```
leszko, line 49: Could you update this description? This image is not only for OpenShift/Kubernetes.

hazelcast-enterprise/Dockerfile
```diff
@@ -26,6 +26,10 @@ ARG EUREKA_PLUGIN_URLS='https://repo1.maven.org/maven2/com/hazelcast/hazelcast-e
 
 # Runtime constants / variables
 ENV HZ_HOME="${HZ_HOME}" \
+    HZ_CP_MOUNT="${HZ_HOME}/external" \
```
leszko, line 29: How about making these env similar to `hazelcast-oss`? Like not to specify `10001`?

hazelcast-enterprise/Dockerfile
```diff
@@ -35,32 +39,44 @@ ENV HZ_HOME="${HZ_HOME}" \
...
+    && sed -i 's/^enabled=.*/enabled=0/g' /etc/dnf/plugins/subscription-manager.conf \
+    && dnf config-manager --disable \
+    && dnf -y update-minimal --security --sec-severity=Important --sec-severity=Critical --setopt=tsflags=nodocs \
+    && echo "Installing new packages" \
+    && dnf -y --setopt=tsflags=nodocs install java-11-openjdk &> /dev/null \
     && echo "Downloading Hazelcast and related JARs" \
-    && mkdir "${HZ_HOME}/lib" \
     && cd "${HZ_HOME}/lib" \
-    && for JAR_URL in ${HAZELCAST_ALL_URL} ${CACHE_API_URL} ${PROMETHEUS_AGENT_URL} ${NETTY_URL} ${NETTY_TCNATIVE_URL} ${EUREKA_PLUGIN_URLS} ${LOG4J2_URLS}; do curl -sf -O -L ${JAR_URL} || exit $?; done \
+    && for JAR_URL in ${HAZELCAST_ALL_URL} ${CACHE_API_URL} ${PROMETHEUS_AGENT_URL} ${NETTY_URL} ${NETTY_TCNATIVE_URL} ${LOG4J2_URLS}; do curl -sf -O -L ${JAR_URL}; done \
```
leszko, line 67: Do we plan to have the Eureka plugin in? I see you removed it from here but didn't remove `ARG EUREKA_PLUGIN_URLS`. I'm fine with removing from Hazelcast Enterprise image, but could you please make sure none of our customers uses Eureka atm?

hazelcast-enterprise/Dockerfile
```diff
@@ -35,32 +39,44 @@ ENV HZ_HOME="${HZ_HOME}" \
...
 
 COPY *.sh *.yaml *.jar *.properties ${HZ_HOME}/
+COPY licenses /licenses
 
 # Install
-RUN echo "Installing new APK packages" \
-    && apk add --no-cache openjdk11-jre bash apr curl procps nss \
-    && echo "Installing old OpenSSL 1.0 from Alpine 3.8" \
-    && apk add --no-cache --repository http://dl-cdn.alpinelinux.org/alpine/v3.8/main/ openssl=1.0.2u-r0 \
+RUN mkdir -p "${HZ_HOME}/lib" "$HZ_CP_MOUNT" \
```
leszko, line 58: Do we need this `HZ_CP_MOUNT`? I'd keep the directories structure similar to `hazelcast-oss`.

hazelcast-enterprise/Dockerfile
```diff
@@ -35,32 +39,44 @@ ENV HZ_HOME="${HZ_HOME}" \
...
-    && for JAR_URL in ${HAZELCAST_ALL_URL} ${CACHE_API_URL} ${PROMETHEUS_AGENT_URL} ${NETTY_URL} ${NETTY_TCNATIVE_URL} ${EUREKA_PLUGIN_URLS} ${LOG4J2_URLS}; do curl -sf -O -L ${JAR_URL} || exit $?; done \
+    && for JAR_URL in ${HAZELCAST_ALL_URL} ${CACHE_API_URL} ${PROMETHEUS_AGENT_URL} ${NETTY_URL} ${NETTY_TCNATIVE_URL} ${LOG4J2_URLS}; do curl -sf -O -L ${JAR_URL}; done \
     && mv jmx_prometheus_javaagent-*.jar jmx_prometheus_javaagent.jar \
-    && echo "Granting read permission to ${HZ_HOME}" \
-    && chmod -R +r ${HZ_HOME} \
-    && echo "Setting Pardot ID to 'docker'" \
-    && echo 'hazelcastDownloadId=docker' > "${HZ_HOME}/hazelcast-download.properties" \
-    && echo "Cleaning APK packages" \
-    && rm -rf /var/cache/apk/*
+    && echo "Adding non-root user" \
```
leszko, line 69: Could you make this part (responsible for adding new user) similar to what we have in `hazelcast-oss`?

hazelcast-enterprise/Dockerfile
```diff
@@ -35,32 +39,44 @@ ENV HZ_HOME="${HZ_HOME}" \
...
+    && dnf -y --setopt=tsflags=nodocs install java-11-openjdk &> /dev/null \
     && echo "Downloading Hazelcast and related JARs" \
-    && mkdir "${HZ_HOME}/lib" \
     && cd "${HZ_HOME}/lib" \
-    && for JAR_URL in ${HAZELCAST_ALL_URL} ${CACHE_API_URL} ${PROMETHEUS_AGENT_URL} ${NETTY_URL} ${NETTY_TCNATIVE_URL} ${EUREKA_PLUGIN_URLS} ${LOG4J2_URLS}; do curl -sf -O -L ${JAR_URL} || exit $?; done \
+    && for JAR_URL in ${HAZELCAST_ALL_URL} ${CACHE_API_URL} ${PROMETHEUS_AGENT_URL} ${NETTY_URL} ${NETTY_TCNATIVE_URL} ${LOG4J2_URLS}; do curl -sf -O -L ${JAR_URL}; done \
     && mv jmx_prometheus_javaagent-*.jar jmx_prometheus_javaagent.jar \
-    && echo "Granting read permission to ${HZ_HOME}" \
-    && chmod -R +r ${HZ_HOME} \
-    && echo "Setting Pardot ID to 'docker'" \
```
leszko, line 55: We still need pardot ID I guess.

leszko (commented):
Added my comments. I think this change is very good in terms of maintenance. However, we need to take a careful approach. See my comments inline and below.

From the technical part, I think you could make the Dockerfile as similar to `hazlecast-oss` (or former `hazelcast-enterprise`) as possible. 

There is also the business-technical part of this. There are some non-backward-compatible implications of this change and I'd suggest creating a TDD describing them.

1. We remove Eureka plugin, we need to make sure none of our customers uses it.
2. The image size is bigger
3. The base image is Red Hat based. Technically, it should be fine, but are there any legal restrictions? Or the license for UBI is the same open as Alpine?
4. You need access to the Red Hat registry in order to build the image. Are any of our customers who build their own images and may have issues with that?
5. What are tools do for their enterprise images (like Redis, Mongo, etc.). Do they all use UBI images?

The other aspect is the performance, but we actually never checked it for the Alpine-based image, so I guess we can skip this for now.

So, I think we need a TDD and I'd like to get the review from at least @jerrinot and @huseyinbabal (Cloud uses our current Docker Hub Enterprise image).

hazelcast-enterprise/Dockerfile
kwart: If we are updating the lib, we should switch to even newer version `4.1.53.Final` and the latest tcnative: `2.0.34.Final`. They are currently used in the hazelcast-enterprise `master` branch.

kwart (commented):
I think the unification of enterprise images (EE+openshift) could be a good step, but we need to handle it carefully.

.travis.yml
hasancelik: I have changed base image from `registry.redhat.io/ubi8/ubi-minimal` to `registry.access.redhat.com/ubi8/ubi-minimal` at the recommendation of @kwart so creds are not required anymore.

hazelcast-enterprise/Dockerfile
hasancelik: All Netty dependencies were already removed at master branch so all of them will be gone when the PR rebased.

hazelcast-enterprise/Dockerfile
hasancelik: Actually, we used it wrong all the time, it should be image/hazelcast version. And yes it is mandatory, there are tests for it and other labels at `Certification Build`. Here is the related section from [official doc](https://connect.redhat.com/zones/containers/container-certification-policy-guide#version_label_exists):
Screen Shot 2020-12-16 at 17 28 37

hazelcast-enterprise/Dockerfile
hasancelik: Done 👍

hazelcast-enterprise/Dockerfile
hasancelik: Done 👍

hazelcast-enterprise/Dockerfile
hasancelik: Done 👍

hazelcast-enterprise/Dockerfile
hasancelik: Actually, I didn't have such a plan for this PR, it just a copy paste error. I want to remove it with another PR but I think we  can do it now 👍 I will talk with support team.

hazelcast-enterprise/Dockerfile
hasancelik: 👍

hazelcast-enterprise/Dockerfile
hasancelik: Done 👍

hazelcast-enterprise/Dockerfile
```diff
@@ -1,11 +1,11 @@
-FROM alpine:3.11
+FROM registry.redhat.io/ubi8/ubi
```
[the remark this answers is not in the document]
hasancelik: both of them are done 👍

hazelcast-enterprise/Dockerfile
```diff
@@ -35,32 +39,44 @@ ENV HZ_HOME="${HZ_HOME}" \
...
-    && apk add --no-cache openjdk11-jre bash apr curl procps nss \
-    && echo "Installing old OpenSSL 1.0 from Alpine 3.8" \
-    && apk add --no-cache --repository http://dl-cdn.alpinelinux.org/alpine/v3.8/main/ openssl=1.0.2u-r0 \
+RUN mkdir -p "${HZ_HOME}/lib" "$HZ_CP_MOUNT" \
+    && echo "Disabling subscription-manager plugin to prevent redundant logs" \
+    && sed -i 's/^enabled=.*/enabled=0/g' /etc/dnf/plugins/subscription-manager.conf \
+    && dnf config-manager --disable \
+    && dnf -y update-minimal --security --sec-severity=Important --sec-severity=Critical --setopt=tsflags=nodocs \
+    && echo "Installing new packages" \
+    && dnf -y --setopt=tsflags=nodocs install java-11-openjdk &> /dev/null \
```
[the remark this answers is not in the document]
hasancelik: > Headless JRE should be sufficient (java-11-openjdk-headless).

good point 👍 

> If we keep the netty and tcnative libs in the image we should also install the openssl and apr packages to make it work.

netty and tcnative jars were already removed at master branch.

pivovarit:
@hasancelik what's the updated size after the layers got merged?

hasancelik:
@pivovarit current image size is [129 MB](https://hub.docker.com/layers/hasancelik/hazelcast-enterprise/ubi8-minimal-compress/images/sha256-37ca72a97d8a4e13abafeff6b282073c8b33de875d505dd5a5cdee0ce9fd30e8?context=explore).

pivovarit:
Well, that's nice. 180MB did not sound good

hasancelik:
180 MB also contains eureka jars but yes, current size is better.

hasancelik:
@huseyinbabal and @jerrinot could you review the PR? Here is the [TDD](https://hazelcast.atlassian.net/wiki/spaces/PM/pages/2918645770/Unification+of+RedHat+Connect+and+DockerHub+IMDG+Enterprise+Images+TDD), you can find details about UBI redistribution and licensing at there.

kwart (approved):
LGTM :+1:

jerrinot (approved):
I did not do technical review. I agree with the unification and using UBI. even if it means a somewhat bigger image.

hazelcast-enterprise/Dockerfile
```diff
@@ -30,30 +27,44 @@ ENV HZ_HOME="${HZ_HOME}" \
     CLASSPATH="" \
     JAVA_OPTS=""
 
+LABEL name="Hazelcast IMDG Enterprise" \
+      maintainer="info@hazelcast.com" \
+      vendor="Hazelcast, Inc." \
+      version="${HZ_VERSION}" \
+      release="1" \
+      summary="RHEL UBI based Hazelcast IMDG Enterprise Image" \
```
leszko, line 35: How about changing the summary and description to just: "Hazelcast IMDG Enterprise Image"?

hazelcast-enterprise/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM alpine:3.12.3
+FROM registry.access.redhat.com/ubi8/ubi-minimal
```
leszko, line 1: Shouldn't we have the specific version of the base image here? Otherwise, we risk the following issues:
- If RH pushes by mistake the wrong `latest` image, we have an issue
- Two same `docker build` can result in different results (if `latest` is changed in the meantime)
- Some of the commands below may stop working and some point and we discover it the hard way.

hazelcast-enterprise/Dockerfile
```diff
@@ -14,13 +14,10 @@ ARG HAZELCAST_ALL_URL="https://repository.hazelcast.com/release/com/hazelcast/ha
...
 
 # Runtime variables
 ENV HZ_HOME="${HZ_HOME}" \
+    USER_NAME="hazelcast" \
```
leszko, line 20: Inline `USER_NAME` to move to `ARG`. I guess you don't need it at runtime.

leszko (commented):
Added my comments.

hazelcast-enterprise/Dockerfile
hasancelik: good catch 👍

hazelcast-enterprise/Dockerfile
hasancelik: Actually, their(RedHat Registries) latest tag usage is a little bit different from DockerHub, `latest` tag is like symbolic link at linux. They just change its pointer, do not push separate image for it. But your points are valid, I have changed it to specific version(8.3) 👍

leszko (approved):

huseyinbabal (approved):
At security and operational level, you always discuss lots of this, my main criteria is this image should be carefully tested in k8s level. If integration team is already ok with this, LGTM!

PR #180, opened 2020-10-27
fix agent params for remote debugger on a newer Java

jerrinot:
This is needed in a newer Java which is bundled in our Docker image. 
See: https://bugs.openjdk.java.net/browse/JDK-8175050


hasancelik (approved):
good catch @jerrinot 👍

PR #185, opened 2020-11-03
Fix vulnerabilities

leszko:
Remove OpenSSL and fix other vulnerabilities


kwart (commented):
We should probably remove the OpenSSL and its Java bindings (`netty` + `netty-tcnative` deps)  too. Java 11 improved the performance of its own TLS engine. Using the wrapped OpenSSL engine in Java 11 on the other hand was slower than in Java 8.

leszko:
> We should probably remove the OpenSSL and its Java bindings (`netty` + `netty-tcnative` deps) too. Java 11 improved the performance of its own TLS engine. Using the wrapped OpenSSL engine in Java 11 on the other hand was slower than in Java 8.

What about the customers? Don't they currently use OpenSSL? @sertugkaya

leszko:
As discussed, removed OpenSSL. @kwart @hasancelik PTAL.

kwart (approved):

hasancelik (approved):

PR #186, opened 2020-11-03
Fix vulnerabilities for 3.12.z

leszko:
Remove OpenSSL and fix other vulnerabilities.


hazelcast-enterprise/Dockerfile
```diff
@@ -44,7 +44,9 @@ COPY *.xml *.sh *.yaml *.jar *.properties ${HZ_HOME}/
 
 # Install
 RUN echo "Installing new APK packages" \
-    && apk add --no-cache openjdk8-jre bash apr openssl curl procps nss \
+    && apk add --no-cache openjdk11-jre bash apr curl \
+    && echo "Installing old OpenSSL 1.0 from Alpine 3.8" \
+    && apk add --no-cache --repository http://dl-cdn.alpinelinux.org/alpine/v3.8/main/ openssl=1.0.2u-r0 \
```
kwart, line 49: Let's get rid of this and provide just Java's own TLS engine out-of-the-box.

leszko: Hmm, TBH I'm slightly afraid to remove OpenSSL in the patch release. Are we sure no customer is using it? @sertugkaya

sertugkaya: @leszko @kwart we see increased adoption of OpenSSL over TLS/SSL in Java, as that's something we suggest to customers for encryption in IMDG as well. But I guess that's not you're asking. 
I'm not sure how changing above setting will affect existing users as it's only about the installation step, right? Since everybody will have TLS/SSL in Java but not everybody might have OpenSSL in their env what harm would this make?

leszko: It's not only about installation steps. It's our Docker image. So, if we remove OpenSSL here, then any customer using Hazelcast Enterprise Docker image won't be able to use OpenSSL.

However, the question is: do we have any customer who uses Hazelcast Enterprise in the form of Docker image and uses OpenSSL at the same time?

sertugkaya: In that case I would suggest to keep it and I guess there are some customers using already. Regardless, since both OpenSSL and Docker images are being more adapted each day, any customer can decide to use it tomorrow as we officially support it: https://docs.hazelcast.org/docs/latest/manual/html-single/#integrating-openssl-boringssl

leszko: In that case, I think we should keep it and just document the vulnerability? WDYT @kwart ?

kwart: I don't like it!
When it's about security, it should not be resolved by "documenting the vulnerability".
Even now, we do [nasty repo-magic](https://github.com/hazelcast/hazelcast-docker/blob/v4.1/hazelcast-enterprise/Dockerfile#L46-L47) to get outdated OpenSSL to our Docker image  (workarounding the https://github.com/netty/netty-tcnative/issues/518).

The OpenSSL [is not enabled by default](https://github.com/hazelcast/hazelcast-enterprise/blob/v4.1/hazelcast-enterprise/src/main/java/com/hazelcast/internal/nio/ssl/AbstractTLSChannelInitializer.java#L76-L91) on Java 11 (and newer).  OpenSSL was significantly slower on Java 11 (see [internal tests](https://hazelcast.atlassian.net/wiki/spaces/SEC/pages/1771438276/TLS+performance+tests+2019)).

If we really want to keep an OOTB native TLS engine in the enterprise image, we could switch to the statically linked BoringSSL. It would mean switching from the musl-based Alpine base image to another one. ( IIRC the BoringSSL doesn't work on Alpine.) See also #177

leszko:
As discussed, removed OpenSSL. @kwart @hasancelik PTAL.

hasancelik (approved):
👍

kwart (approved):

PR #189, opened 2020-11-17
Bump Anchore action version.

enozcan:
The current version seems quite old and just started to fail due to deprecated GHA properties:
https://github.com/hazelcast/hazelcast-docker/actions/runs/367236703
This is also why the build for this commit (6c838e9) fails.


leszko (approved):
👍

hasancelik (approved):

Issue #195, opened 2020-11-24
Image is not using latest JDK11 version


cangencer:
Docker image is using JDK 11.0.5, latest is 11.0.9

pivovarit:
Just one note, if we switched the base image to `adoptopenjdk/openjdk11:jdk-11.0.9_11-alpine`, we could leverage Dependabot and simplify our _Dockerfile_.

ldziedziul:
Done within https://github.com/hazelcast/hazelcast-docker/pull/576

PR #199, opened 2020-11-24
[PoC] Simplify base images by depending on Adoptopenjdk's Alpine-based image

pivovarit:
Related: https://github.com/hazelcast/hazelcast-docker/issues/195

### DIFF
--- a/hazelcast-enterprise/Dockerfile
+++ b/hazelcast-enterprise/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.12.1
+FROM adoptopenjdk/openjdk11:jdk-11.0.9_11-alpine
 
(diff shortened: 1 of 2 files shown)


cangencer:
Should we not use JDK15 ? it has much faster startup time (see https://cl4es.github.io/2020/12/06/Towards-OpenJDK-17.html)

pivovarit:
I think we should but I made an assumption that we want to fix the bug first, and migrate to newer JDK as a separate effort

pivovarit:
I think we should, but I created this one just to spark a discussion about the base image

pivovarit:
386MB vs 225MB, I think it's not worth it.

PR #204, opened 2020-12-02
Default to JDK's default Garbage Collector with -XX:MaxGCPauseMillis=5

pivovarit:
closes: https://github.com/hazelcast/hazelcast-docker/issues/197
closes: https://github.com/hazelcast/hazelcast-docker/issues/196

PR #205, opened 2020-12-09
Use openjdk15-jre as runtime

pivovarit:
Not sure if `--repository=http://dl-cdn.alpinelinux.org/alpine/edge/testing` is acceptable for production-ready artifacts, but it looks like we'd need to wait a bit until it's promoted from _testing_ to _community_

### DIFF
--- a/hazelcast-enterprise/Dockerfile
+++ b/hazelcast-enterprise/Dockerfile
@@ -37,7 +37,9 @@ COPY *.sh *.yaml *.jar *.properties ${HZ_HOME}/
 
 # Install
 RUN echo "Installing new APK packages" \
-    && apk add --no-cache openjdk11-jre bash curl \
+    && apk update \
+    && apk add --no-cache openjdk15-jre --repository=http://dl-cdn.alpinelinux.org/alpine/edge/testing \
+    && apk add --no-cache bash curl \
     && echo "Downloading Hazelcast and related JARs" \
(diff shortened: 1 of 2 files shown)


hazelcast-enterprise/Dockerfile
```diff
@@ -37,7 +37,9 @@ COPY *.sh *.yaml *.jar *.properties ${HZ_HOME}/
 
 # Install
 RUN echo "Installing new APK packages" \
-    && apk add --no-cache openjdk11-jre bash curl \
+    && apk update \
+    && apk add --no-cache openjdk15-jre --repository=http://dl-cdn.alpinelinux.org/alpine/edge/testing \
```
leszko, line 41: Do we want to use the `testing` repository? Seems like something not production-ready. Plus, what are the benefits for upgrading to JRE15?

pivovarit:
> ERROR: http://dl-cdn.alpinelinux.org/alpine/v3.12/main: network error (check Internet connection and firewall)

Unrelated build failures. Works locally

hazelcast-enterprise/Dockerfile
pivovarit: That's literally what I asked in the PR description :)

G1 was greatly improved in JDK15.

pivovarit: https://github.com/hazelcast/hazelcast-docker/pull/204#issuecomment-737737676

PR #209, opened 2020-12-16
Separate push jobs from security scans

pivovarit:
Image pushing and image scanning should be separate to not misguide people into thinking that push failed if it's just a scan

image


hasancelik (approved):
LGTM 👍 👏

PR #213, opened 2020-12-29
Use config override at README examples

hasancelik:
`TLS_ENABLED` -- > `HZ_NETWORK_SSL_ENABLED `
`HZ_LICENSE_KEY` --> `HZ_LICENSEKEY`
`JAVA_OPTS="-Dhazelcast.local.publicAddress` --> `HZ_NETWORK_PUBLICADDRESS`

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,13 +18,13 @@ This command will pull a Hazelcast Docker image and run a new Hazelcast instance
 For the simplest end-to-end scenario, you can create a Hazelcast cluster with two Docker containers and access it from the client application.
 
 ```
-$ docker run -e JAVA_OPTS="-Dhazelcast.local.publicAddress=<host_ip>:5701" -p 5701:5701 hazelcast/hazelcast:$HAZELCAST_VERSION
-$ docker run -e JAVA_OPTS="-Dhazelcast.local.publicAddress=<host_ip>:5702" -p 5702:5701 hazelcast/hazelcast:$HAZELCAST_VERSION
+$ docker run -e HZ_NETWORK_PUBLICADDRESS=<host_ip>:5701 -p 5701:5701 hazelcast/hazelcast:$HAZELCAST_VERSION
+$ docker run -e HZ_NETWORK_PUBLICADDRESS=<host_ip>:5702 -p 5702:5701 hazelcast/hazelcast:$HAZELCAST_VERSION
 ```
 
 Note that:
 * each container must publish the `5701` port under a different host machine port (`5701` and `5702` in the example)
-* supplying a custom `hazelcast.local.publicAddress` is critical for autodiscovery. Otherwise, Hazelcast will bind to Docker's internal ports.
+* supplying a custom `HZ_NETWORK_PUBLICADDRESS` is critical for autodiscovery. Otherwise, Hazelcast will bind to Docker's internal ports.
 * `<host_ip>` needs to be the host machine address that will be used for the Hazelcast communication
 
 After setting up the cluster, you can start the [client](https://github.com/hazelcast/hazelcast-code-samples/tree/master/clients/basic) application to check if it works correctly.
@@ -36,7 +36,7 @@ You can launch a Hazelcast Enterprise Docker Container by running the following 
 Please request trial license [here](https://hazelcast.com/hazelcast-enterprise-download/) or contact sales@hazelcast.com.
 
 ```
-$ docker run -e HZ_LICENSE_KEY=<your_license_key> hazelcast/hazelcast-enterprise:$HAZELCAST_VERSION
+$ docker run -e HZ_LICENSEKEY=<your_license_key> hazelcast/hazelcast-enterprise:$HAZELCAST_VERSION
 ```
 
 ### Hazelcast Enterprise Hello World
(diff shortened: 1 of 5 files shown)


pivovarit (approved):

leszko (commented):
Is this change backward-compatible?

hasancelik:
> Is this change backward-compatible?

@leszko config override introduced with 4.1 release so `HZ_NETWORK_SSL_ENABLED, HZ_LICENSEKEY, HZ_NETWORK_PUBLICADDRESS` are not avaliable for previous releases like 4.0.x . As a result, it's not backward-compatible.

leszko:
> > Is this change backward-compatible?
> 
> @leszko config override introduced with 4.1 release so `HZ_NETWORK_SSL_ENABLED, HZ_LICENSEKEY, HZ_NETWORK_PUBLICADDRESS` are not avaliable for previous releases like 4.0.x . As a result, it's not backward-compatible.

Then I guess we need to tweak it, so that the old env variables would still work correctly.

pivovarit:
We could just provide two versions? one for >4.1 and one for 4.0.x?

leszko:
> We could just provide two versions? one for >4.1 and one for 4.0.x?

What do you mean by 2 versions? I think we need to check the old variable and if it's defined then assign it to the new variable. For the sake of backward-compatibility.

pivovarit:
Hmm, now I see you asked about something else. This should definitely stay backward compatible.

I just meant explanations in docs that could be misleading for older images (since there's no support for the env-based config)

hasancelik:
@leszko @pivovarit What about creating a new branch for 4.0.z images?

leszko:
@hasancelik why would we need such a branch?

About the docs, I think the docs at master should be about the current master version. If someone needs docs for older images, then they can check the tagged README.

My point about the backward-compatibility was just that these commands should both work now.
```
$ docker run -e HZ_LICENSE_KEY=<license-key> hazelcast/hazelcast-enterprise
$ docker run -e HZ_LICENSEKEY=<license-key> hazelcast/hazelcast-enterprise
```

hasancelik:
> About the docs, I think the docs at master should be about the current master version. If someone needs docs for older images, then they can check the tagged README.

@leszko with next `4.0.x` release, HZ_VERSION will be updated to `4.0.4` and related image/tag will be created from master branch by IMDG release script. This tag includes same README file with `4.1.1` so release management process of docker images will become a little bit complicated. If we decide to continue with only master, we need to provide two command/explanation for 4.0.x and 4.1.x at README.

leszko:
> > why would we need such a branch?
> 
> > About the docs, I think the docs at master should be about the current master version. If someone needs docs for older images, then they can check the tagged README.
> 
> @leszko with next `4.0.x` release, HZ_VERSION will be updated to `4.0.4` and related image/tag will be created from master branch by IMDG release script. This tag includes same README file with `4.1.1` so release management process of docker images will become a little bit complicated. If we decide to continue with only master, we need to provide two command/explanation for 4.0.x and 4.1.x at README.

Ok, then, yes, we should create the branch you suggested.

README.md
```diff
@@ -132,6 +107,34 @@ $ docker run -m 512M hazelcast/hazelcast:$HAZELCAST_VERSION
...
+hazelcast:
+  cluster-name: dev
+  network:
+    port:
+      auto-increment: true
+      port-count: 100
+      port: 5701
+    join:
+      auto-detection:
+        enabled: true
```
leszko, line 125:
```suggestion
hazelcast:
  cluster-name: dev
  network:
    port:
      auto-increment: true
      port-count: 100
      port: 5701
```

Auto-detection is enabled by default, so it does not make much sense to use it in the example.

README.md
```diff
@@ -132,6 +107,34 @@ $ docker run -m 512M hazelcast/hazelcast:$HAZELCAST_VERSION
...
+```
+
+If you want to use the environment variables, the above would be represented as a set of the following environment variables:
+```bash
+$ docker run -e HZ_CLUSTERNAME=dev \
+  -e HZ_NETWORK_PORT_AUTOINCREMENT=true \
+  -e HZ_NETWORK_PORT_PORTCOUNT=100 \
+  -e HZ_NETWORK_PORT_PORT=5701 \
+  -e HZ_NETWORK_JOIN_AUTODETECTION_ENABLED=true \
+  hazelcast/hazelcast
```
leszko, line 135:
```suggestion
$ docker run -e HZ_CLUSTERNAME=dev \
  -e HZ_NETWORK_PORT_AUTOINCREMENT=true \
  -e HZ_NETWORK_PORT_PORTCOUNT=100 \
  -e HZ_NETWORK_PORT_PORT=5701 \
  hazelcast/hazelcast
```

README.md
```diff
@@ -152,6 +155,40 @@ $ docker run -e CLASSPATH="/opt/hazelcast/CLASSPATH_EXT/*" -v /home/ubuntu/hazel
 
 Alternatively, you can [extend Hazelcast base image](#extending-hazelcast-base-image) adding your custom JARs.
 
+
+### Using TLS (Hazelcast Enterprise Only)
+
+The `HZ_NETWORK_SSL_ENABLED` environment variable can be used to enable TLS for the communication. You can check TSL related section at the documantation, [here](https://docs.hazelcast.org/docs/4.1.1/manual/html-single/index.html#integrating-openssl-boringssl).
```
leszko, line 161: I guess what you describe here (and in point 1) is related to OpenSSL, which we actually don't recommend. I guess point 2 should be enough to use TLS with the standard Java TLS implementation.

hasancelik: yes, you are totally right, done 👍

leszko (approved):
LGTM 👍 

Sorry for such a late review.

@pivovarit could you please look and approve as well?

pivovarit (approved):
LGTM

PR #218, opened 2021-01-07
Oss snapshot release

mtyazici:
This PR changes the Hazelcast IMDG Docker image release structure. There will be two workflows.
- `latest_snapshot_push.yml` will be triggered by the Github REST API, its input is `HZ_VERSION`, which will be a snapshot version. The image created by this workflow will have the tag `latest-snapshot`.
- `tag_image_push.yml` will be triggered when a new tag is created whose name starts with `v`. If the created tag's version is greater than the rest of the tags, the created image's tags will be `latest` and `x.y.z`. Otherwise it will only have the tag `x.y.z`.

To support installing snapshot versions of `hazelcast-all` I added a script named `get-hz-all-version.sh` which will return the url of `hazelcast-all.jar` depending on if the `HZ_VERSION` is a snapshot.

There are some minor changes in the `hazelcast-oss` Dockerfile, these changes are all related to snapshot version support.

NOTE: Currently QEMU used on Github Actions gives an error when `Java`, incidentally `maven`, is used in the Dockerfile on architectures `ppc64le` and `s390x`. Instead of using maven in the Dockerfile to install snapshot versions, I wrote a script named `get-hz-all-url.sh`, it returns the URL for both snapshot and released versions of `hazelcast-all`. However, this is a intermediate solution and will be replaced when the problem on QEMU is solved. There is an issue created in Alpine repository but it is mentioned the problem is related to QEMU, https://gitlab.alpinelinux.org/alpine/aports/-/issues/12128.


hazelcast-oss/get-hz-all-url.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+if [[ "${HZ_VERSION}" == *"SNAPSHOT"* ]]
+then
+    xml=$(curl -sSL https://oss.sonatype.org/content/repositories/snapshots/com/hazelcast/hazelcast-all/${HZ_VERSION}/maven-metadata.xml)
+    version=$(echo $xml | xpath -q  -e  '/metadata/versioning/snapshotVersions/snapshotVersion[1]/value' | grep -oP '(?<=<value>).*(?=</value>)')
```
leszko, line 6:
```suggestion
    version=$(echo $xml | xpath -q -e '/metadata/versioning/snapshotVersions/snapshotVersion[1]/value' | grep -oP '(?<=<value>).*(?=</value>)')
```

leszko, line 6: You always select the first value, are these snapshot versions always sorted and the first one is the latest one?

hazelcast-oss/get-hz-all-url.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
leszko, line 1: You can add some comments here about the reasoning behind this script. Mention for sure that:
- we need to extract the version manually (not with Maven), since `mvn` doesn't work with `dockerx` emulation
- how you extract the snapshot value (that's it's always the first entry in the XML file)

.github/workflows/tag_image_push.yml
```diff
@@ -28,37 +30,52 @@ jobs:
...
-
       - name: Login to Docker Hub
         run: echo "${{ secrets.DOCKER_PASSWORD }}" | docker login -u "${{ secrets.DOCKER_USERNAME }}" --password-stdin
 
       - name: Build/Push OSS image
         run: |
-          docker buildx build --push \
-            --tag hazelcast/hazelcast:${{ env.RELEASE_VERSION }} \
-            --platform=linux/arm64,linux/amd64,linux/ppc64le,linux/s390x hazelcast-oss
+          if [[ "${{ env.PUSH_LATEST }}" == "yes" ]]; then
```
leszko, line 38: I think you can simplify it to something like:
```
TAGS="-tag hazelcast/hazelcast:${{ env.RELEASE_VERSION }}"
if [[ "${{ env.PUSH_LATEST }}" == "yes" ]]; then
    TAGS="${TAGS} --tag hazelcast/hazelcast:latest"
fi
docker buildx build --push \
              --build-arg HZ_VERSION=${{ env.RELEASE_VERSION }} \
              ${TAGS} \
              --platform=linux/arm64,linux/amd64,linux/ppc64le,linux/s390x hazelcast-oss
```

.github/workflows/tag_image_push.yml
```diff
@@ -28,37 +30,52 @@ jobs:
...
+            docker buildx build --push \
+              --build-arg HZ_VERSION=${{env.RELEASE_VERSION }} \
+              --tag hazelcast/hazelcast:${{ env.RELEASE_VERSION }} \
+              --tag hazelcast/hazelcast:latest \
+              --platform=linux/arm64,linux/amd64,linux/ppc64le,linux/s390x hazelcast-oss
+          else
+            docker buildx build --push \
```
leszko, line 45: So this push does not change the `latest` tag, am I correct?

.github/workflows/latest_snapshot_push.yml
```diff
@@ -1,12 +1,14 @@
 name: Build latest OS and EE image
 
 on:
-  push:
-    branches:
-      - master
+  workflow_dispatch:
```
leszko, line 4: Who will trigger the snapshot build?

leszko (commented):
Added my comments, but it would be good if @hasancelik and @pivovarit review it as well.

hazelcast-oss/get-hz-all-url.sh
mtyazici: `maven-metadata.xml` only holds the value for the latest snapshot version. The list consist of `jar`, `pom`, `sources` and `javadoc`. So it doesn't really matter which element of the list I get.

hazelcast-oss/get-hz-all-url.sh
mtyazici: Sure, I will add comments to the script.

.github/workflows/tag_image_push.yml
mtyazici: That is a good idea, I will change it. Thanks :smile:

.github/workflows/tag_image_push.yml
mtyazici: Yes, it doesn't affect the `latest` tag.

.github/workflows/latest_snapshot_push.yml
mtyazici: After this PR passes, @hasancelik and I are planning to do it together. It will be triggered by Jenkins release pipeline.

hasancelik: Afaik, it should be triggered by `Hazelcast-4.master-deploy` jenkins job, not release pipeline.

.github/workflows/latest_snapshot_push.yml
```diff
@@ -1,12 +1,14 @@
...
+    inputs:
+      HZ_VERSION:
+        description: 'Version of Hazelcast to build the image for'
+        required: true
 
 jobs:
-  push:
+  build:
```
hasancelik, line 11: `push` is more reasonable, I think :)

.github/workflows/tag_image_push.yml
```diff
@@ -2,23 +2,25 @@ name: Build OS and EE image
...
   push:
-    branches:
-      - "!*"
     tags:
-      - "v4.*"
+      - "v*"
 
 jobs:
-  push:
+  build:
```
hasancelik, line 9: same :)

hasancelik (changes requested):
LGTM except few minor comments 👍

.github/workflows/latest_snapshot_push.yml
mtyazici: I will change it :)

.github/workflows/tag_image_push.yml
mtyazici: Will change this one too.

.github/workflows/tag_image_push.yml
```diff
@@ -2,23 +2,25 @@ name: Build OS and EE image
 
 on:
   push:
-    branches:
-      - "!*"
     tags:
-      - "v4.*"
```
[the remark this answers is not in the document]
mtyazici: I will revert the changes

hasancelik (approved):

.github/workflows/tag_image_push.yml
```diff
@@ -28,37 +32,44 @@ jobs:
...
+          fi
           docker buildx build --push \
-            --tag hazelcast/hazelcast:${{ env.RELEASE_VERSION }} \
+            --build-arg HZ_VERSION=${{ env.RELEASE_VERSION }} \
+            ${TAGS} \
             --platform=linux/arm64,linux/amd64,linux/ppc64le,linux/s390x hazelcast-oss
+
       - name: Build/Push EE image
         run: |
+          TAGS="--tag hazelcast/hazelcast:${{ env.RELEASE_VERSION }}"
```
leszko, line 51: I guess it should be `hazelcast-enterprise`

.github/workflows/tag_image_push.yml
```diff
@@ -28,37 +32,44 @@ jobs:
...
+          if [[ "${{ env.PUSH_LATEST }}" == "yes" ]]; then
+            TAGS="${TAGS} --tag hazelcast/hazelcast:latest"
```
leszko, line 53: I think it should be `hazelcsat/hazelcast-enterprise`

leszko (commented):
Added two comments.

Also, could you please update the name of the PR, now it's `Oss snapshot release", but it seems it's not only related to OSS (but also to enterprise).

.github/workflows/tag_image_push.yml
mtyazici: Yes you are right, when I was testing I commented out these parts. I will update them.

.github/workflows/tag_image_push.yml
mtyazici: Answered above.

mtyazici:
> Also, could you please update the name of the PR, now it's `Oss snapshot release", but it seems it's not only related to OSS (but also to enterprise).
@leszko I also need to change the Dockerfile for enterprise version. But it is also being changed by Hasan. So, with this PR only `hazelcast-oss` snapshots will be released and enterprise version will fail. Should I still change the title?

hazelcast-oss/Dockerfile
```diff
@@ -36,23 +36,25 @@ COPY *.sh *.yaml *.jar *.properties ${HZ_HOME}/
 
 # Install
 RUN echo "Installing new APK packages" \
-    && apk add --no-cache openjdk11-jre bash curl \
+    && apk add --no-cache openjdk11-jre bash curl perl-xml-xpath \
     && echo "Downloading Hazelcast and related JARs" \
     && mkdir "${HZ_HOME}/lib" \
     && cd "${HZ_HOME}/lib" \
+    && HAZELCAST_ALL_URL=$(${HZ_HOME}/get-hz-all-url.sh) \
```
pivovarit, line 43: Would be better to define an environment variable by using a dedicated keyword:

https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#env

leszko (approved):

.github/workflows/latest_snapshot_push.yml
mtyazici: We changed the triggerin Jenkins job to  `Hazelcast-docker-snapshot-release`.

hazelcast-oss/Dockerfile
pivovarit: Also, you should not need to `$(${HZ_HOME}/get-hz-all-url.sh)`, it should be enough to just:

> HAZELCAST_ALL_URL="${HZ_HOME}/get-hz-all-url.sh"

pivovarit:
Two minor notes from me, but most important problems were spotted already and solved.

hazelcast-oss/Dockerfile
mtyazici: I actually tried running `ENV HAZELCAST_ALL_URL=$(${HZ_HOME}/get-hz-all-url.sh)`. However the ENV variable's value turned out to be `$(/opt/hazelcast/get-hz-all-url.sh)`. So I don't think we can run commands in the `ENV` step.

Even if we could run commands in `ENV` step, there is also one downside to using `ENV` as a seperate step because that would split the installation `RUN` step into two parts. I am not sure how much bigger the image size would get but I believe it would be significant.

pivovarit: Ah, wait. You're running the script and not defining the environment variable. Ignore my comment :)

pivovarit (approved):

PR #228, opened 2021-01-21
Hz Enterprise Red Hat release

mtyazici:
A new workflow `tag_image_push_rhel.yml` is added. The workflow will be triggered by the same way as `tag_image_push.yml` and it does the following:
- It finds the created tag's version then builds the Docker image
- It pushes the Docker image to Red Hat scan registry and waits for scan to be completed
- After scanning finishes, it creates a project in the OCP environment.
- It deploys a new cluster of 3 members and 1 management center using the Hazelcast EE image from the scan registry.
- It verifies the cluster size.
- It deletes the created project.
- If cluster is formed it publishes the Hazelcast EE image in the scan registry.


.github/workflows/tag_image_push_rhel.yml
```diff
@@ -0,0 +1,119 @@
+name: hz-enterprise-operator-rhel-autorelease
+
+on:
+  push:
+    branches:
+      - "!*"
+    tags:
+      - "v4.*"
```
hasancelik, line 8: @mtyazici I could not see any push command for `latest` tag. What happens if we  release patch release(4.0.4) after minor release(4.1.1)?
For example, Dockerhub release script pushes latest tag seperately:
https://github.com/hazelcast/hazelcast-docker/blob/master/.github/workflows/tag_image_push.yml#L42-L45

mtyazici: Currently we have no way of changing latest tag from code. It has to be done manually. I will create a ticket about it and share the results with you.

mtyazici: I created the case but it is going to be resolved much later than we anticipated. I will also create an issue about it in this repository.

hasancelik (approved):

PR #230, opened 2021-01-25
replace hazelcast-download.properties at hz-all-jar with modified one

hasancelik:
Since 3.11.6 image, docker phone home records have not been saved properly. They were recorded as a `maven` ping, it means that containers uses `hazelcast-download.properties` file at jar itself, not classpath one. Still I can not find the main cause of this issue but overriding the file at jar with file that is prepared by Dockerfile does fix the issue.

back-port PR --> https://github.com/hazelcast/hazelcast-docker/pull/231

### DIFF
--- a/hazelcast-enterprise/Dockerfile
+++ b/hazelcast-enterprise/Dockerfile
@@ -18,7 +18,6 @@ ARG LOG4J2_URLS="https://repo1.maven.org/maven2/org/apache/logging/log4j/log4j-c
 
 # Runtime variables
 ENV HZ_HOME="${HZ_HOME}" \
-    HZ_VERSION="${HZ_VERSION}" \
     CLASSPATH_DEFAULT="${HZ_HOME}/*:${HZ_HOME}/lib/*" \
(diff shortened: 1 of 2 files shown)


hazelcast-oss/Dockerfile
```diff
@@ -21,6 +21,7 @@ ARG EUREKA_PLUGIN_URLS='https://repo1.maven.org/maven2/com/hazelcast/hazelcast-e
 
 # Runtime variables
 ENV HZ_HOME="${HZ_HOME}" \
+    HZ_VERSION="${HZ_VERSION}" \
```
enozcan, line 24: Could it be resolved without setting env? Isn't ARG enough to refer `hazelcast-all-${HZ_VERSION}.jar`?

hasancelik: good point, removed 👍

leszko (approved):
Thanks for the fix @hasancelik

hazelcast-enterprise/Dockerfile
```diff
@@ -44,18 +44,20 @@ COPY licenses /licenses
...
     && mv jmx_prometheus_javaagent-*.jar jmx_prometheus_javaagent.jar \
+    && echo "Setting Pardot ID to 'docker'" \
+    && echo 'hazelcastDownloadId=docker' > "hazelcast-download.properties" \
+    && zip -f hazelcast-enterprise-all-${HZ_VERSION}.jar hazelcast-download.properties \
     && echo "Granting read permission to ${HZ_HOME}" \
     && chmod -R +r ${HZ_HOME} \
-    && echo "Setting Pardot ID to 'docker'" \
-    && echo 'hazelcastDownloadId=docker' > "${HZ_HOME}/hazelcast-download.properties" \
-    && echo "Removed cached package data" \
+    && echo "Removed cached package data and unnecessary tools" \
```
enozcan, line 59: nit:
```suggestion
    && echo "Removing cached package data and unnecessary tools" \
```

enozcan (approved):

PR #243, opened 2021-02-18
Add filter for tags with 'BETA' keyword in tag_image_push.yml

mtyazici:
This prevents images with `BETA` tags to be published with `latest` tag.


hasancelik (approved):
👍

PR #250, opened 2021-03-15
Add conditionals for operations specific to workflows in master branch

mtyazici:
With these changes when these workflows are triggered in other branches, they will not do the operations specific to the master branch.


pivovarit (approved):

hasancelik (approved):
💯

PR #274, opened 2021-06-23
Added warning about HZ_LICENSE_KEY env variable for 3.x and 4.x users

hasancelik:
- [ ] Update docker commands at hazelcast doc:
    https://docs.hazelcast.com/imdg/4.2/installation/installing-using-docker.html


leszko:
Should we keep the `HZ_LICENSE_KEY` variable in 5.0 for the sake of backwards-compatibility or we're fine to ask the customers to change the env variable to `HZ_LICENSEKEY`? WDYT @mmedenjak @jerrinot @cangencer

PR #287, opened 2021-07-06
fix Trivy vulnerabilities + removed Anchore scan [CN-109]

SeriyBg:


.github/containerscan/allowedlist.yaml
```diff
@@ -3,6 +3,9 @@ general:
     # netty-tcnative still requires libssl 1.0
     - CVE-2019-1543
     - CVE-2020-1967
```
hasancelik, line 5: I think we can remove this CVEs from here because OpenSSL related JARs were already removed by https://github.com/hazelcast/hazelcast-docker/pull/185

hasancelik (approved):
LGTM 👍

.github/containerscan/allowedlist.yaml
SeriyBg: Done

PR #298, opened 2021-08-04
Enable ignore-unfixed option for Trivy

hasancelik:
Additional changes:
- use official trivy Github Action

Trivy steps at EE workflows are currently failing after I changed severity level to medium(its default valut is high):
https://github.com/hazelcast/hazelcast-docker/runs/3241439215?check_suite_focus=true
Snyk steps are not failing because Snyk ignores unfixed(won't fix) ones as a default behaviour:
https://support.snyk.io/hc/en-us/articles/360000923498-How-can-I-ignore-a-vulnerability- 

@leszko I checked most of medium vulns are reported by Trivy were marked as `will not fix` by RedHat so checking them one by one every time then adding into whitelist.yaml to make our builds are green again is just time-consumer. If vulnerability can not be fixed, it can be invisible for us. WDYT?

### DIFF
--- a/.github/workflows/ee_latest_snapshot_push.yml
+++ b/.github/workflows/ee_latest_snapshot_push.yml
@@ -49,12 +49,12 @@ jobs:
             https://api.github.com/repos/hazelcast/charts/actions/workflows/push-hazelcast-enterprise-snapshot.yml/dispatches \
             -d '{"ref":"master"}'
 
-      - name: Scan Hazelcast Enterprise image by Azure (Trivy + Dockle)
-        if: always()
-        uses: Azure/container-scan@v0
+      - name: Scan OSS image by Trivy
+        uses: aquasecurity/trivy-action@master
         with:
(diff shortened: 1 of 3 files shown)


leszko:
Hi @jerrinot wdyt?

As far as I understand there are just too many things we'd need to check manually and add to the whitelist, which is time-consuming. Do you have any idea for it?

hasancelik:
Hi @jerrinot, what is your opinion?

hasancelik:
It seems Snyk changed their policy about unfixed vulns:
```
 In addition, for Red Hat Enterprise Linux we will now show both fixed and unfixed CVEs, whereas the RHSA only shows fixed CVEs. For Amazon Linux, we will still only support fixed CVEs at this time. 
```
https://updates.snyk.io/ --> `Snyk Container Security Data Update to Red Hat Enterprise Linux (RHEL) and Amazon Linux`

frant-hartm:
This has been done in https://github.com/hazelcast/hazelcast-docker/pull/373

PR #306, opened 2021-09-28
[CN-209] Merge 5.0 based changes:

hasancelik:
- all enhancements from 5.0-snapshot branch like multiple tag support and version updates
- modify existing github actions with multi tag support


SeriyBg (changes requested):
One small comment, other than that LGTM

.github/workflows/4-1-z-backport_vulnerability_scan copy.yml
```diff
@@ -0,0 +1,57 @@
+name: Scheduled 4.1.z Vulnerability Scan
```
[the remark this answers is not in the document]
hasancelik: good catch 👍 🙂

SeriyBg (approved):

.github/workflows/tag_image_push.yml
```diff
@@ -57,18 +67,19 @@ jobs:
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
-          repository: hazelcast/hazelcast
-          short-description: Hazelcast IMDG Docker Image
+          repository: ${DOCKER_ORG}/hazelcast
```
mtyazici, line 70: I don't think the actions can receive the environment variables in shell style. I believe the line should be as follows
```suggestion
          repository: ${{ env.DOCKER_ORG }}/hazelcast
```

.github/workflows/tag_image_push.yml
```diff
@@ -78,8 +89,8 @@ jobs:
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
-          repository: hazelcast/hazelcast-enterprise
-          short-description: Hazelcast IMDG Enterprise Docker Image
+          repository: ${DOCKER_ORG}/hazelcast-enterprise
```
mtyazici, line 92: Same as above
```suggestion
          repository: ${{ env.DOCKER_ORG }}/hazelcast-enterprise
```

mtyazici (changes requested):
I left two comments and I will mention another one here, other than that LGTM :+1: 

The file `hazelcast-enterprise/values.yaml` you are deleting here is used by one workflow, instead of deleting it here in another PR we can move it in `.github` folder and update the workflow .

.github/workflows/tag_image_push.yml
hasancelik: Actually, it is working but you are right we should follow convention 👍

hasancelik:
@mtyazici I updated the PR, PTAL

mtyazici (approved):
:+1:

PR #328, opened 2021-12-03
Remove lib/* from CLASSPATH

### LINKED ISSUE #290: 5.x docker doesn't report downloadId correctly
looking at phone homes - I only see Maven or Distribution.

frant-hartm:
This is already added by the `hz-start` script.
Adding this to CLASSPATH results in the lib/* directory present 2x on
the final classpath.

Accidentally it is put before the default classpath, causing a change in
the loading order. Incorrect hazelcast-download.properties is loaded and
phone home reports as 'maven'.

Fixed #290


hasancelik (approved):

ldziedziul (approved):
good finding!

PR #344, opened 2022-02-08
Add custom distribution zip to gitignore

frant-hartm:
This is useful when trying local build or when we use the docker
repository as submodule of the main repository.


ldziedziul:
👍

PR #346, opened 2022-02-14
Update logging configuration after changes in distribution

frant-hartm:
In the Docker image, we use a specific config file without the rolling file
appender (only log to console). The changes in the main distribution
need to be applied to these config files as well.

Set default value for LOGGING_LEVEL in log4j2 properties.

In https://github.com/hazelcast/hazelcast/pull/20528 we removed setting
the default value in the scripts and moved it to the log4j2 config file.

### DIFF
--- a/hazelcast-oss/log4j2.properties
+++ b/hazelcast-oss/log4j2.properties
@@ -4,6 +4,7 @@ appender.console.type=Console
 appender.console.name=STDOUT
 appender.console.layout.type=PatternLayout
 appender.console.layout.pattern=${env:LOGGING_PATTERN}
+appender.console.layout.alwaysWriteExceptions=false
 
-rootLogger.level=${env:LOGGING_LEVEL}
+rootLogger.level=${env:LOGGING_LEVEL:-INFO}
 rootLogger.appenderRef.stdout.ref=STDOUT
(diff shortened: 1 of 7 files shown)


ldziedziul (approved):
Looks good. What do you think about adding info about `LOGGING_CONFIG=log4j2-json.properties` to the README.md?

ldziedziul (approved):
Nice!

hazelcast-oss/log4j2.properties
```diff
@@ -4,6 +4,7 @@ appender.console.type=Console
 appender.console.name=STDOUT
 appender.console.layout.type=PatternLayout
 appender.console.layout.pattern=${env:LOGGING_PATTERN}
+appender.console.layout.alwaysWriteExceptions=false
 
-rootLogger.level=${env:LOGGING_LEVEL}
+rootLogger.level=${env:LOGGING_LEVEL:-INFO}
```
SeriyBg, line 9: Is the `-` a typo, or is it a supported syntax?

SeriyBg (approved):
One small question. Other than that LGTM

hazelcast-oss/log4j2.properties
frant-hartm: It is the syntax for the default value - see https://logging.apache.org/log4j/2.x/manual/lookups.html.

SeriyBg: Oh, I see. I thought the syntax was the same as for spring, that is `${LOGGING_LEVEL:INFO}.`

PR #373, opened 2022-04-19
Improve vulnerability scans [HZ-1090]

frant-hartm:
- report only high and critical vulnerabilities (less noise, we
investigate and fix all high & critical to have green build)
- exclude base image vulnerabilities using `--exclude-base-image-vulns`
- ignore unfixed using `TRIVY_IGNORE_UNFIXED: true` - we usually can't
do anything about them
- whitelist false positives
- whitelist vulnerabilities we already track in main repo

### DIFF
--- a/.github/workflows/vulnerability_scan.yml
+++ b/.github/workflows/vulnerability_scan.yml
@@ -6,6 +6,8 @@ on:
   push:
     branches:
       - master
+  pull_request:
+    types: [ opened, synchronize, edited ]
 
 jobs:
   build:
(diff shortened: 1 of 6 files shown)


ldziedziul (approved):

.github/workflows/vulnerability_scan.yml
```diff
@@ -26,6 +26,8 @@ jobs:
       - name: Scan OSS image by Azure (Trivy + Dockle)
         if: always()
         uses: Azure/container-scan@v0
+        env:
+          TRIVY_IGNORE_UNFIXED: true
```
ldziedziul, line 30: Awesome ❤️

kwart (approved):
I'm a little bit afraid that we might miss a base image update due to the skipping. Still, the smaller area properly tested and handled is better than a larger area with ignored failures.

frant-hartm:
@kwart I think we will get this via Trivy - we have `TRIVY_IGNORE_UNFIXED: true`, but once an issue is fixed it will get reported. That's how I discovered that we still used redhat minimal image 8.4 for 4.1.z/4.2.z

PR #409, opened 2022-05-31
Update vulnerable eureka dependencies [4.2.z]

frant-hartm:
Updated dependencies: jackson and gson


ldziedziul (approved):
updated deps: jackson and gson

PR #445, opened 2022-09-14
Fix UBI repository ids

frant-hartm:
This page https://access.redhat.com/articles/4238681 lists the repositories with the `-rpms` suffix.

Not sure if this changed recently or if there is some other change causing the previous value not to work.


ldziedziul (approved):
3 days ago or less: https://webcache.googleusercontent.com/search?q=cache:FPaz5O0AAiYJ:https://access.redhat.com/articles/4238681&cd=1&hl=en&ct=clnk&gl=en

kwart (approved):

PR #453, opened 2022-09-26
Use LANG instead of LC_ALL and set it on the OS side too

kwart:
Backports #348 to `5.1.z`.


TomaszGaweda (approved):
"emojis-not-displayed" Loving it :D

PR #470, opened 2022-11-07
No fail fast and secrets passing to subflow

TomaszGaweda:


.github/workflows/vulnerability_scan_subworkflow.yml
```diff
@@ -11,12 +11,15 @@ on:
             ref:
                 required: true
                 type: string
+        secrets:
+            SNYK_TOKEN:
+                required: true
```
ldziedziul, line 16: could we use `${{ secrets.SNYK_TOKEN }}` for `workflow_dispatch` ?

TomaszGaweda: not sure about it, this method is mentioned in GitHub docs

zhmaeff: https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow
AFAICS, we can use secrets param only for `workflow_call`.
For `workflow_dispatch`, secrets are available as per org/repo settings.

ldziedziul (approved):
Nice job!

PR #472, opened 2022-11-09
Split workflow into build and scan

TomaszGaweda:


frant-hartm (changes requested):
What issue is this PR trying to address?

ldziedziul:
> What issue is this PR trying to address?

To make more visible if the build failure comes from the `build` steps or `scan` steps (security vulnerabilities), although after considerdation I think it won't bring any benefit and make everything more complicated

PR #522, opened 2022-11-28
Add simple smoke test for OSS Dockerfile

ldziedziul:
This PR adds workflow testing OSS Dockerfile using our CLC application.

Example run: https://github.com/ldziedziul/hazelcast-docker/actions/runs/3565826014/jobs/5991476430

### DIFF
--- a/.github/scripts/simple-smoke-test.sh
+++ b/.github/scripts/simple-smoke-test.sh
@@ -27,7 +27,7 @@ function test_docker_image() {
 }
 
 function install_clc() {
-  CLC_URL="https://github.com/hazelcast/hazelcast-commandline-client/releases/download/${CLC_VERSION}/hazelcast-clc_${CLV_VERSION}_linux_amd64.tar.gz"
+  CLC_URL="https://github.com/hazelcast/hazelcast-commandline-client/releases/download/${CLC_VERSION}/hazelcast-clc_${CLC_VERSION}_linux_amd64.tar.gz"
   curl -L $CLC_URL | tar xzf - --strip-components=1 -C /usr/local/bin
   chmod +x /usr/local/bin/clc
 }
--- a/.github/scripts/simple-smoke-test.sh
+++ b/.github/scripts/simple-smoke-test.sh
@@ -27,7 +27,7 @@ function test_docker_image() {
 }
 
 function install_clc() {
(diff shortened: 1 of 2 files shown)


.github/scripts/simple-smoke-test.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+
+set -e
+set -o pipefail
+
+export CLV_VERSION=v5.2.0-beta3-preview1
```
frant-hartm, line 6: CLV_VERSION?

.github/scripts/simple-smoke-test.sh
```diff
@@ -0,0 +1,36 @@
...
+
+function test_docker_image() {
+    local image=$1
+    local container_name=$2
+    echo "Starting container '$container_name' from image '$image'"
+    docker run -it --name "$container_name" -d -p5701:5701 "$image"
```
frant-hartm, line 12: I am not sure where this is stored but maybe add `--rm` so the container is removed after being stopped.

frant-hartm (changes requested):

.github/scripts/simple-smoke-test.sh
ldziedziul: I've skipped '--rm' to be able to get logs and store them as artifacts. GitHub runners are cleared before each reuse

.github/scripts/simple-smoke-test.sh
```diff
@@ -0,0 +1,36 @@
...
+    docker stop "$container_name"
+
+    if [ "$expected" != "$actual" ]; then
+        echo "Expected to read '${expected}' but got '${actual}'"
+        exit 1;
+    fi
+}
+
+function install_clc() {
+  CLC_URL="https://github.com/hazelcast/hazelcast-commandline-client/releases/download/${CLV_VERSION}/hazelcast-clc_${CLV_VERSION}_linux_amd64.tar.gz"
```
frant-hartm, line 30: Here as well.
```suggestion
  CLC_URL="https://github.com/hazelcast/hazelcast-commandline-client/releases/download/${CLC_VERSION}/hazelcast-clc_${CLV_VERSION}_linux_amd64.tar.gz"
```

frant-hartm (approved):

PR #554, opened 2023-04-14
Update docker tags for minor and major versions

ldziedziul:
Adds dynamic labels for major (e.g. hazelcast/hazelcast:5) and minor (e.g. hazelcast/hazelcast:5.2) versions, when pushing the patch version (e.g. hazelcast/hazelcast:5.2.3).

Examples valid today:
```
hazelcast/hazelcast:5 → hazelcast/hazelcast:5.2.3
hazelcast/hazelcast:4 → hazelcast/hazelcast:4.2.7
```

Needs to be backported to all supported version branches

Fixes https://hazelcast.atlassian.net/browse/HZ-1382

Tested by running the workflow against `5.1.5` version:
image

Support for multiple tags for RHEL images will be added in a separate PR as there's no clear way how to do it with RH Openshift API. I've submitted a case to RH Support

### DIFF
--- a/.github/scripts/get-tags-to-push_tests.sh
+++ b/.github/scripts/get-tags-to-push_tests.sh
@@ -39,7 +39,7 @@ log_header "Tests for get_tags_to_push"
 assert_tags_to_push "5.2.0" "5.2.0"
 assert_tags_to_push "5.2.1" "5.2.1"
 assert_tags_to_push "5.1.99" "5.1.99 5.1"
-assert_tags_to_push "5.99.0" "5.99.0 5.99 5 latest" #this needs adjustment when we release 6.0
+assert_tags_to_push "4.99.0" "4.99.0 4.99 4"
 assert_tags_to_push "99.0.0" "99.0.0 99.0 99 latest"
 assert_tags_to_push "5.3.0-BETA-1" "5.3.0-BETA-1"
 assert_tags_to_push "5.99.0-BETA-1" "5.99.0-BETA-1"


.github/scripts/get-tags-to-push.sh
```diff
@@ -0,0 +1,38 @@
...
+
+if verlte "$LATEST_FOR_MAJOR" "$VERSION_TO_RELEASE"; then
+   TAGS_TO_PUSH+=($MAJOR_VERSION_TO_RELEASE)
+fi
+
+if verlte "$LATEST" "$VERSION_TO_RELEASE"; then
+   TAGS_TO_PUSH+=(latest)
+fi
+
+echo "${TAGS_TO_PUSH[@]}"
```
frant-hartm, line 38: Could we add a test for this file? I tried couple of examples and it mostly works ok, but e.g. for 

```
$ ./get-tags-to-push.sh 5.3.0-BETA-3
5.3.0-BETA-3 5.3 5 latest
```

I think tags `5` and `latest` shouldn't be pushed for beta, and maybe 5.3 also, but not sure there.

frant-hartm (changes requested):
Will this part of the task:

> Supported Hazelcast Images should be rebuilt whenever the underlying base image is updated (or if this is not possible for any reason, then rebuild regularly e.g. weekly) ;

be done in a separate PR?

ldziedziul:
> Will this part of the task:
> 
> > Supported Hazelcast Images should be rebuilt whenever the underlying base image is updated (or if this is not possible for any reason, then rebuild regularly e.g. weekly) ;
> 
> be done in a separate PR?

Yes

.github/scripts/get-tags-to-push.sh
ldziedziul: > Could we add a test for this file?

Added https://github.com/hazelcast/hazelcast-docker/pull/554/commits/938969d0570ec6c4e11f0feeb84bd1ca62afb0d9

> I think tags 5 and latest shouldn't be pushed for beta

I agree, fixed: https://github.com/hazelcast/hazelcast-docker/pull/554/commits/f48f084672b497aeaa514df0c58910d778fc4fc1

frant-hartm (approved):

PR #587, opened 2023-05-25
Remove azure container scan

ldziedziul:
Replace not maintained Azure/containerscan with two actions for Trivy and Dockle

Test run: https://github.com/hazelcast/hazelcast-docker/actions/runs/5079024434/jobs/9124252301


.github/containerscan/.trivyignore
```diff
@@ -0,0 +1 @@
+# Add ignored CVEs in new lines below
```
frant-hartm, line 1: We can add CVE-2022-3171 with reference to https://github.com/hazelcast/hazelcast/issues/20807 and https://issues.apache.org/jira/browse/HADOOP-18197

.github/containerscan/trivy.yaml
```diff
@@ -0,0 +1,8 @@
+dependency-tree: true
+exit-code: 1
+ignorefile: .github/containerscan/.trivyignore
+severity:
+  - CRITICAL
+  - HIGH
+vulnerability:
+  ignore-unfixed: true
```
frant-hartm, line 8: nit:
```suggestion
  ignore-unfixed: true

```

frant-hartm (approved):

PR #643, opened 2023-08-25
Build NLC images for arm64

ldziedziul:


atilsensalduz (approved):

orcunc (approved):

.github/workflows/ee-nlc-snapshot-push.yml
```diff
@@ -73,5 +81,11 @@ jobs:
           path: |
             ${{ env.DOCKER_LOG_FILE_EE }}
 
-      - name: Push EE image
-        run: docker push ${NLC_IMAGE_NAME}:${HZ_VERSION} 
+      - name: Build/Push EE image
+        run: |
+          docker buildx build --push \
```
frant-hartm, line 86: The image gets built at this line for x86

https://github.com/hazelcast/hazelcast-docker/pull/643/files#diff-08145d3368b04a30d18de9633da71d473424b0d738f463ba799c204f11660c80R57-R61

and then again here for both architectures. Is the second build cached?

If not could we use buildx to
-  build, but not push
-  run the smoke test
- push image

Also does it make sense to run the smoke test for both architectures?

frant-hartm (changes requested):

.github/workflows/ee-nlc-snapshot-push.yml
ldziedziul: @frant-hartm  after this change https://github.com/hazelcast/hazelcast-docker/pull/643/commits/b087e6cc16552e13dc482f3c72b19d2f5bc0fc87 we use cache (https://github.com/hazelcast/hazelcast-docker/actions/runs/5997599199/job/16264281270): 
image

> Also does it make sense to run the smoke test for both architectures?

We would need to complicate the pipeline and build each architecture independently and then test it as it's not allowed to store (and of course run) arm64 images on the amd64 machine. 

We use the same approach (test amd64, push all) in all other workflows using buildx. If you want to revisit that I would suggest to make it a separate task.

frant-hartm (approved):

PR #667, opened 2023-08-31
Fix finding latest patch version. [5.3.z]

ldziedziul:
Backport of https://github.com/hazelcast/hazelcast-docker/pull/666

'5.4.0-DEVEL-5' was recognized as the latest '4.0' patch version because it contains '4.0'. It resulted with build failures:

image

Within this fix we check if a tested version starts with a given minor version


ldziedziul:
But we need it to pass the PR tests

frant-hartm (approved):

PR #673, opened 2023-10-09
Update EE image to ubi 9.3

ldziedziul:
Deprecated functionality in 9.3:

https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html-single/9.3_release_notes/index#deprecated-functionality


frant-hartm (approved):
Fine to upgrade on master.

On the maintenance branches, I think we should delay the upgrade till the next patch release, so we don't make such a significant change for already existing images.

PR #674, opened 2023-10-23
Do not push multiple tags for DEVEL versions

ldziedziul:
It also breaks automatic rebuilds of the `latest` tag


frant-hartm (approved):
Thanks!

PR #675, opened 2023-10-23
Fix 5.3.4 deployment by pushing 5.3.5

nishaatr:
Unfortunately, 5.3.4 was partially released incorrectly and we cannot revert or restrict as its deployed to Maven Central
We are going to release 5.3.5 instead which is based off 5.3.3 and skip 5.3.3 altogether
This should make 5.3.5 as the latest release and users should be able to pick it up overshadowing 5.3.4

Fixes: https://hazelcast.atlassian.net/browse/HZ-3546


JackPGreen (approved):

Patras3 (approved):
LGTM.

PR #710, opened 2024-01-10
Fix upgrade to ubi 9.3

ldziedziul:
Make all microdnf commands non-interactive

Fixes https://github.com/hazelcast/hazelcast-docker/pull/673

The PR above passed tests since there were no packages to upgrade when the PR checks were run

### DIFF
--- a/hazelcast-enterprise/Dockerfile
+++ b/hazelcast-enterprise/Dockerfile
@@ -39,7 +39,7 @@ COPY *.jar get-hz-ee-dist-zip.sh hazelcast-*.zip ${HZ_HOME}/
 
 # Install
 RUN echo "Installing new packages" \
-    && microdnf update --nodocs \
+    && microdnf -y update --nodocs \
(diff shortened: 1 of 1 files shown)


Patras3 (approved):

TomaszGaweda:
OSS uses e.g.:
```
 && apk upgrade --no-cache \
```

I think we should add `-y` also there

ldziedziul:
> OSS uses e.g.:
> 
> ```
>  && apk upgrade --no-cache \
> ```
> 
> I think we should add `-y` also there

@TomaszGaweda have you notice problem with the current command? I would prefer keeping commands simple if they simply work

TomaszGaweda:
No, but it may strike back like microdnf command in the future

TomaszGaweda (approved):

TomaszGaweda:
I won't block the PR, just IMHO worth considering

PR #715, opened 2024-01-22
Use a single definition of the supported JDK versions

ldziedziul:
Define supported JDKs in a single place: `./.github/workflows/get-supported-jdks.yaml`


JackPGreen (approved):
I like this, a lot.

orcunc (approved):

Issue #753, opened 2024-06-11
Do we need `suffix` paremeter?


JackPGreen:
Do we need `suffix` paremeter?

_Originally posted by @ldziedziul in https://github.com/hazelcast/hazelcast-docker/pull/747#discussion_r1603284092_
            

The `suffix` parameter in the scripts always duplicates `hz_version`, with `-` prepended if non-empty.

This duplication should be removed.

PR #754, opened 2024-06-13
Remove `suffix` parameter from scripts

JackPGreen:
`suffix` is a hardcoded derivative of `variant`, and can instead be derived dynamically.

`tag_image_push_rhel.yml` used `suffix`, but never actually set it (unlike `tag_image_push.yml` which did), so hardcoded it to `""` instead.

Fixes: https://github.com/hazelcast/hazelcast-docker/issues/753

### DIFF
--- a/.github/workflows/tag_image_push_rhel.yml
+++ b/.github/workflows/tag_image_push_rhel.yml
@@ -42,6 +42,10 @@ jobs:
       fail-fast: false
       matrix:
         jdk: ${{ fromJSON(needs.jdks.outputs.jdks) }}
+        variant: [ 'slim','' ]
+        include:
+          - variant: slim
+          - variant: ''
     steps:
       - name: Set HZ version as environment variable
         run: |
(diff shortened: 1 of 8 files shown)


ldziedziul (commented):
Am I missing something or `matrix.suffix` is not defined and always empty in `tag_image_push_rhel.yml`?

JackPGreen:
> Am I missing something or `matrix.suffix` is not defined and always empty in `tag_image_push_rhel.yml`?

Glad it's not just me.

JackPGreen:
> > Am I missing something or `matrix.suffix` is not defined and always empty in `tag_image_push_rhel.yml`?
> 
> Glad it's not just me.

Yes, [it's empty](https://github.com/hazelcast/hazelcast-docker/actions/runs/9502129469/job/26189379828):
`TAGS_TO_PUSH=$(get_tags_to_push 5.4.0 "" "17" "$DEFAULT_JDK")`

.github/workflows/tag_image_push_rhel.yml
```diff
@@ -42,6 +42,10 @@ jobs:
       fail-fast: false
       matrix:
         jdk: ${{ fromJSON(needs.jdks.outputs.jdks) }}
+        variant: [ 'slim','' ]
```
ldziedziul, line 45: Why have you added it?

We haven't push slim versions to OpenShift Catalog

JackPGreen: I _assumed_ it was an oversight as it’s in the non-RHEL job. If that’s wrong I can take it out instead and update the params to suit?

ldziedziul: There's no request for it so I would keep it as it is (i.e. no slim version on RHEL)

JackPGreen: bff66f46b7260e7f3b25124010093e23ec1eec79

ldziedziul (approved):

PR #755, opened 2024-06-14
Do not create latest-snapshot tag for OS docker image [DI-150]

ldziedziul:
It doesn't make sense to keep this tag anymore, probably we won't publish any OS snapshot docker images

### DIFF
--- a/.github/workflows/oss_latest_snapshot_push.yml
+++ b/.github/workflows/oss_latest_snapshot_push.yml
@@ -94,9 +94,6 @@ jobs:
           . .github/scripts/oss-build.functions.sh
           
           VERSIONS=("$HZ_VERSION")
-          if [[ "${{ github.ref }}" == "refs/heads/master" ]]; then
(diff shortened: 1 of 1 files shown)


JackPGreen (approved):

nishaatr (changes requested):
What happens if someone builds against master?
Or is not possible?

ldziedziul:
It's possible (if one triggers the workflow manually or within our release pipeline). In this case the workflow will push `5.5.0-SNAPSHOT` tag but won't create `latest-snapshot`

nishaatr:
> It's possible (if one triggers the workflow manually or within our release pipeline). In this case the workflow will push `5.5.0-SNAPSHOT` tag but won't create `latest-snapshot`

understood. makes sense
But should we allow the Action to run manually on `master` whether intentionally or unintentionally?  
I don't think Release Pipeline triggers Docker for master, and as for Jenkins, we have made the changes not to publish OSS master

PR #757, opened 2024-06-19
Fix flaky CLC installation in tests

ldziedziul:
Use specific version as the downloading the latest version often fails: 

```
Hazelcast CLC Installer (c) 2023 Hazelcast, Inc.

INFO  Creating the Home directory: /home/runner/.hazelcast
ERROR could not determine the latest version
Error: Process completed with exit code 1.
```

See also: https://hazelcast.slack.com/archives/C0319N7HV8W/p1718023112329759

### DIFF
--- a/.github/scripts/simple-smoke-test.sh
+++ b/.github/scripts/simple-smoke-test.sh
@@ -35,7 +35,9 @@ function test_docker_image() {
 }
 
 function install_clc() {
-  curl https://hazelcast.com/clc/install.sh | bash
+  # Use specific version as the downloading the latest version often fails: https://hazelcast.slack.com/archives/C0319N7HV8W/p1718023112329759
+  local CLC_VERSION=v5.4.0
+  curl https://raw.githubusercontent.com/hazelcast/hazelcast-commandline-client/main/extras/unix/install.sh | bash -s -- --version "$CLC_VERSION"
   export PATH=$PATH:$HOME/.hazelcast/bin
   clc config add default cluster.name=dev cluster.address=localhost
 }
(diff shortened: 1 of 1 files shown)


JackPGreen (changes requested):
I don't like this:
- who is going to maintain `CLC_VERSION` as the release increases?
- if the released version doesn't work, it's good were catching it (albeit in the wrong place)


I understand this causes downstream builds to fail but I think that's a symptom, not an issue to be worked around.

ldziedziul:
> I don't like this:
> 
> * who is going to maintain `CLC_VERSION` as the release increases?
> 
> I understand this causes downstream builds to fail but I think that's a symptom, not an issue to be worked around.

This is only a temporary workaround until CLC team will fix it, I'm not happy about it either but we shouldn't suffer because of it. 

> > * if the released version doesn't work, it's good were catching it (albeit in the wrong place)

It's not our responsibility, to be testers of CLC. Of course it's nice to test it as a side effect but it should effect our build pipelines. 

The alternative would be to add retry mechanism instead:
```shell
 while ! curl https://hazelcast.com/clc/install.sh | bash
    do
      echo "Retrying..."
      sleep 3
    done
```

What do you think?

JackPGreen:
> > I don't like this:
> > 
> > * who is going to maintain `CLC_VERSION` as the release increases?
> > 
> > I understand this causes downstream builds to fail but I think that's a symptom, not an issue to be worked around.
> 
> This is only a temporary workaround until CLC team will fix it, I'm not happy about it either but we shouldn't suffer because of it.
> 
> > > * if the released version doesn't work, it's good were catching it (albeit in the wrong place)
> 
> It's not our responsibility, to be testers of CLC. Of course it's nice to test it as a side effect but it should effect our build pipelines.
> 
> The alternative would be to add retry mechanism instead:
> 
> ```shell
>  while ! curl https://hazelcast.com/clc/install.sh | bash
>     do
>       echo "Retrying..."
>       sleep 3
>     done
> ```
> 
> What do you think?

I much prefer the second option.

PR #759, opened 2024-06-20
Use private OS distribution ZIPs in preference to public ones [DI-95] [5.4.z]

JackPGreen:
Backport of https://github.com/hazelcast/hazelcast-docker/pull/747

`SNAPSHOT` OS distribution ZIPs (e.g. [those indexed here](https://oss.sonatype.org/content/repositories/snapshots/com/hazelcast/hazelcast-distribution/5.4.0-SNAPSHOT)) will no longer be generated, which causes problems with the `hazelcast-docker` PR builder (amongst others).

Instead, the distribution ZIPs will be uploaded to a private AWS S3 bucket, which requires authentication.

**This was a non-trivial change**:
- Updated the default distribution in the Docker images to be the latest released version
   - This is guaranteed to be available (without any credentials)
- Allows the `Dockerfile` to be executed normally outside of a build (e.g. for the security scan)
- "latest released" is looked up via Maven to ensure this doesn't need any explicit maintenance in future
- Added unit tests to all modified shell scripts
- Moved `get_hz_dist_zip` into `os_build.functions` (and same for EE)
- Created additional `maven.functions` script to support resolution of "latest released" for a given artifact
   - This has to be duplicated for OS & EE because of Docker limitations
- Added `xmllint` to all the runner images that build anything as required for the Maven stuff
- Updated the `get_dist` OS script to generate a pre-signed S3 URL to get it from S3 rather than a Maven repository
- Because this requires authentication, had to move this from inside the image to outside to access the repository secrets
- Simplified the `Dockerfile` arguments - the URL resolution is now being done outside of the `Dockerfile` more of the time, meaning not always required
   - removed the redundant `HZ_VARIANT` parameter - removed `HZ_VERSION` from OS and reduced usage in EE

For testing, I've been doing what I can locally to assert the results between `master` and my branch are the same:
- E.G. the `build-pr.yml` doesn't pass locally for me on EE, I presume due to some `act` incompatibility - so a milestone for my PR is to get the same failure in the same place
- I've build the images locally (outside of `act`, just a plain `docker buildx build hazelcast-os`) and they build ok.
- Because of the reliance on repository secrets, it's not possible to run many tests properly

Fixes: [DI-95](https://hazelcast.atlassian.net/browse/DI-95)

[DI-95]: https://hazelcast.atlassian.net/browse/DI-95?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ldziedziul (approved):
Good work!

Issue #763, opened 2024-06-21
Dependabot PRs failing due to inaccessible AWS secrets


JackPGreen:
[PRs raised by Dependabot fail the PR builder due to being unable to access AWS secrets](https://github.com/hazelcast/hazelcast-docker/pull/762).

[When run manually they pass](https://github.com/hazelcast/hazelcast-docker/actions/runs/9609103495), suggesting a problem with secret accessibility.

The [suggested solution](https://github.com/dependabot/dependabot-core/issues/3253#issuecomment-852541544) is to use `pull_request_target`.

PR #764, opened 2024-06-21
Make PR builder use `pull_request_target` to support Dependabot

JackPGreen:
Fixes https://github.com/hazelcast/hazelcast-docker/issues/763

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -2,7 +2,7 @@ name: Build PR
 
 on:
   workflow_dispatch:
-  pull_request:
+  pull_request_target:
 
 env:
   test_container_name_oss: hazelcast-oss-test


JackPGreen:
Needs further thought.

nishaatr (approved):

ldziedziul:
That's not a solution. Simple switching to `pull_request_target` would make the workflow to test the code from `master` branch and not from the PR. So we wouldn't test PR changes before the merge. We would end up eventually with a broken `master`

I've asked Josef to enable AWS secrets for dependabot user, it should work now

ldziedziul (changes requested):
I think we should close it

JackPGreen:
> I've asked Josef to enable AWS secrets for dependabot user, it should work now

Perfect!

PR #766, opened 2024-06-24
Reduce duplication in scripts between test and released images - `HAZELCAST_ZIP_URL` [DI-156]

JackPGreen:
The generation of the `HAZELCAST_ZIP_URL` parameter should be saved to ensure consistency between images used for testing, and the image that's ultimately deployed to ensure the test is as representative as possible.

In addition, a test has been added to check the type of distribution in the image - EE or OSS - as a sanity check.

Fixes: [DI-156](https://hazelcast.atlassian.net/browse/DI-156)

### DIFF
--- a/.github/scripts/simple-smoke-test.sh
+++ b/.github/scripts/simple-smoke-test.sh
@@ -5,7 +5,8 @@ set -o pipefail
 
 function test_docker_image() {
     local image=$1
-    local container_name=$2    local expected_distribution_type=$3
+    local container_name=$2
+    local expected_distribution_type=$3
 
     if [ "$(docker ps --all --quiet --filter name="$container_name")" ]; then
       echo "Removing existing '$container_name' container"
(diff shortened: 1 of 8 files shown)


.github/scripts/simple-smoke-test.sh
```diff
@@ -6,12 +6,27 @@ set -o pipefail
...
     local image=$1
     local container_name=$2
+    local expected_distribution_type=$3
 
     if [ "$(docker ps --all --quiet --filter name="$container_name")" ]; then
       echo "Removing existing '$container_name' container"
       docker container rm --force "$container_name"
     fi
 
+    echo "Checking if $image is EE"
```
nishaatr, line 16: [nit] could be simplify (not sure about correct syntax)

```
[[ docker run --rm $image bash -c 'compgen -G lib/*enterprise* ]] && distribution_type="ee" || distribution_type="oss"
 echo "Image type $distribution_type"

```

nishaatr (approved):

.github/scripts/simple-smoke-test.sh
JackPGreen: I don't think that's functionally equivalent - it doesn't log either way.

PR #772, opened 2024-06-28
Fix calculation of the zip url

ldziedziul:
Fixing issues introduced in https://github.com/hazelcast/hazelcast-docker/pull/766


kwart (approved):
Thanks for the quick fix.

PR #774, opened 2024-07-01
Use internal snapshot repository for OSS snapshot images [DI-169]

ldziedziul:
- Use internal snapshot maven repository for getting OSS snapshot distribution zip
- Use internal snapshot docker registry for OSS snapshot images

Fixes https://hazelcast.atlassian.net/browse/DI-169

### DIFF
--- a/.github/scripts/assert.sh/assert.sh
+++ b/.github/scripts/assert.sh/assert.sh
@@ -178,7 +178,7 @@ assert_contain() {
   fi
 
   if [ -z "${haystack##*$needle*}" ]; then
-    log_success "Array [$(join_by "," $haystack)] contains $needle :: $msg" || true
+    log_success "[$(join_by "," $haystack)] contains $needle :: $msg" || true
     return 0
   else
     [ "${#msg}" -gt 0 ] && log_failure "Array [$(join_by "," $haystack)] doesn't contain $needle :: $msg" || true
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -16,7 +16,6 @@ jobs:
     name: Prepare environment
     outputs:
       HZ_VERSION_OSS: ${{ steps.get_oss_vars.outputs.HZ_VERSION_OSS }}
-      HAZELCAST_OSS_ZIP_URL: ${{ steps.get_oss_vars.outputs.HAZELCAST_OSS_ZIP_URL }}
       HZ_VERSION_EE: ${{ steps.get_ee_vars.outputs.HZ_VERSION_EE }}
       HAZELCAST_EE_ZIP_URL: ${{ steps.get_ee_vars.outputs.HAZELCAST_EE_ZIP_URL }}
     steps:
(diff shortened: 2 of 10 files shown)


.github/scripts/assert.sh/assert.sh
JackPGreen, line 1: Rather than changing this, can we fix upstream? This isn't the only project that uses this script.

.github/workflows/build-pr.yml
```diff
@@ -11,54 +11,75 @@ env:
...
+      HZ_VERSION_OSS: ${{ steps.get_oss_vars.outputs.HZ_VERSION_OSS }}
+      HZ_VERSION_EE: ${{ steps.get_ee_vars.outputs.HZ_VERSION_EE }}
+      HAZELCAST_EE_ZIP_URL: ${{ steps.get_ee_vars.outputs.HAZELCAST_EE_ZIP_URL }}
     steps:
       - name: Checkout Code
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
 
+      - name: Install xmllint
```
JackPGreen, line 27: Nit - this should probably go back where it was (i.e. reject a build before you setup the env)

.github/scripts/assert.sh/assert.sh
ldziedziul: Actually it's undo of my previous fix ;) Either way I think maybe it's a good idea to put it in `hazelcast/assert.sh` fork

.github/workflows/build-pr.yml
```diff
@@ -11,54 +11,75 @@ env:
...
+
+  build-pr:
+    runs-on: ubuntu-latest
+    name: Build with default JDK
+    needs: [ prepare ]
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
JackPGreen, line 63: Why do we need to fetch more than the current branch?

.github/workflows/build-pr.yml
```diff
@@ -11,54 +11,75 @@ env:
...
-
+          HZ_VERSION="${{ needs.prepare.outputs.HZ_VERSION_OSS }}"
+          
+          # duplicated block as GH doesn't support passing sensitive data between jobs
+          . .github/scripts/oss-build.functions.sh
+          export HZ_SNAPSHOT_INTERNAL_PASSWORD=${{ secrets.HZ_SNAPSHOT_INTERNAL_PASSWORD }}
+          export HZ_SNAPSHOT_INTERNAL_USERNAME=${{ secrets.HZ_SNAPSHOT_INTERNAL_USERNAME }}
+          HAZELCAST_OSS_ZIP_URL=$(get_hz_dist_zip "" "${HZ_VERSION}")
+          
+          curl -sfL "$HAZELCAST_OSS_ZIP_URL" --output $DOCKER_PATH/hazelcast-distribution.zip;
```
JackPGreen, line 79: Nit - prefer long form arguments (what is `sfL`? Not immediately obvious without reading the docs).

.github/workflows/oss_latest_snapshot_push.yml
```diff
@@ -106,8 +112,8 @@ jobs:
           
           VERSIONS=("$HZ_VERSION")
           
-          DOCKER_DIR=hazelcast-oss
-          IMAGE_NAME=${{ env.DOCKER_ORG }}/hazelcast
+          DOCKER_DIR=hazelcast-oss
```
JackPGreen, line 115: Nit - formatting

.github/scripts/assert.sh/assert.sh
ldziedziul: https://hazelcast.atlassian.net/browse/DI-171

.github/workflows/build-pr.yml
ldziedziul: Fixed: https://github.com/hazelcast/hazelcast-docker/pull/774/commits/231f22d2a3f58bf0e7de77e22b25888752126893

.github/workflows/build-pr.yml
ldziedziul: AFAIR because we need tags for `get-tags-to-push.sh` and `version.functions.sh`

.github/workflows/build-pr.yml
ldziedziul: Fixed https://github.com/hazelcast/hazelcast-docker/pull/774/commits/4c9c9fafaa690b5edfda3aa322e0c5a17b869e78

JackPGreen (approved):

nishaatr (approved):
LGTM although my review was high level and not much on logic. But looks ok to me. 
Couple of nit comments

1. `awk` and `curl` being used in many places. may be create a common function and call that. this way easier to maintain and change params in one place
2. I thought `assert.h` was sourced externally. So change might get overwritten if we ever update the file

JackPGreen:
> 2. I thought `assert.h` was sourced externally. So change might get overwritten if we ever update the file

https://hazelcast.atlassian.net/browse/DI-171

PR #776, opened 2024-07-02
Update `simple-smoke-test` to check Hazelcast version [DI-163]

JackPGreen:
- adds a version check
- refactors `simple-smoke-test`
   - adds functions to aid readability now it's got a bit larger
   - resolve Spellcheck IDE warnings

Fixes: [DI-163](https://hazelcast.atlassian.net/browse/DI-163)

[DI-163]: https://hazelcast.atlassian.net/browse/DI-163?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/scripts/simple-smoke-test.sh
+++ b/.github/scripts/simple-smoke-test.sh
@@ -42,7 +42,7 @@ function check_distribution_type() {
     fi
 
     if [[ "${distribution_type}" != "${expected_distribution_type}" ]]; then
-      echo "Image ${image} should contain ${expected_distribution_type} distribution but ${distribution_type} was detected"
+      echoerr "Image ${image} should contain ${expected_distribution_type} distribution but ${distribution_type} was detected"
       exit 1
     fi
 }
@@ -57,7 +57,7 @@ function check_version() {
     if [[ "${version}" == "${expected_version}" ]]; then
       echo "${image} version identified as ${version}"
     else
-      echo "${image} version was ${version}, not ${expected_version} as expected"
+      echoerr "${image} version was ${version}, not ${expected_version} as expected"
       exit 1
     fi
 }
@@ -83,7 +83,7 @@ function test_map_read_write() {
     docker stop "${container_name}"
 
     if [[ "${expected}" != "${actual}" ]]; then
-        echo "Expected to read '${expected}' but got '${actual}'"
+        echoerr "Expected to read '${expected}' but got '${actual}'"
         exit 1;
     fi
 }
@@ -97,6 +97,12 @@ function install_clc() {
   export PATH=${PATH}:${HOME}/.hazelcast/bin
   clc config add default cluster.name=dev cluster.address=localhost
 }
+
+# Prints the given message to stderr
+function echoerr() {
+  echo "ERROR - $@" 1>&2;
+}
+
 
 install_clc
 test_docker_image "$@"
--- a/.github/scripts/simple-smoke-test.sh
+++ b/.github/scripts/simple-smoke-test.sh
@@ -11,7 +11,7 @@ function test_docker_image() {
     remove_container_if_exists "${container_name}"
 
     check_distribution_type "${image}" "${expected_distribution_type}"
-    check_version "${image}" "${expected_version}"
+    check_image_hz_version "${image}" "${expected_version}"
 
     test_map_read_write "${image}" "${container_name}"
 }
@@ -47,7 +47,7 @@ function check_distribution_type() {
     fi
 }
 
-function check_version() {
+function check_image_hz_version() {
     local image=$1
     local expected_version=$2
 
--- a/.github/scripts/simple-smoke-test.sh
+++ b/.github/scripts/simple-smoke-test.sh
@@ -6,10 +6,25 @@ set -o pipefail
 function test_docker_image() {
     local image=$1
     local container_name=$2
+    local expected_distribution_type=$3
 
     if [ "$(docker ps --all --quiet --filter name="$container_name")" ]; then
       echo "Removing existing '$container_name' container"
       docker container rm --force "$container_name"
+    fi
+
+    echo "Checking if $image is EE"
+    if docker run --rm $image bash -c 'compgen -G lib/*enterprise*'; then
+      echo "EE contents identified"
+      distribution_type="ee"
+    else
+      echo "No EE contents identified - assuming OSS"
+      distribution_type="oss"
+    fi
+
+    if [[ "$distribution_type" != "$expected_distribution_type" ]]; then
+      echo "Distribution was $distribution_type, not $expected_distribution_type as expected"
+      exit 1
     fi
 
     echo "Starting container '$container_name' from image '$image'"
--- a/.github/scripts/simple-smoke-test.sh
+++ b/.github/scripts/simple-smoke-test.sh
@@ -54,8 +54,10 @@ function check_version() {
     echo "Checking ${image} version"
     local version
     version=$(docker run --rm "${image}" bin/hz-cli --version | awk '/Hazelcast/ {print $2}')
-    if [[ "${version}" != "${expected_version}" ]]; then
-      echo "${image} was ${version}, not ${expected_version} as expected"
+    if [[ "${version}" == "${expected_version}" ]]; then
+      echo "${image} version identified as ${version}"
+    else
+      echo "${image} version was ${version}, not ${expected_version} as expected"
       exit 1
     fi
 }
--- a/.github/scripts/simple-smoke-test.sh
+++ b/.github/scripts/simple-smoke-test.sh
@@ -42,7 +42,7 @@ function check_distribution_type() {
     fi
 
     if [[ "${distribution_type}" != "${expected_distribution_type}" ]]; then
-      echo "Distribution was ${distribution_type}, not ${expected_distribution_type} as expected"
+      echo "Image ${image} should contain ${expected_distribution_type} distribution but ${distribution_type} was detected"
       exit 1
     fi
 }
--- a/.github/scripts/simple-smoke-test.sh
+++ b/.github/scripts/simple-smoke-test.sh
@@ -1,20 +1,39 @@
 #!/usr/bin/env bash
 
-set -e
-set -o pipefail
+set -o errexit -o nounset -o pipefail
 
 function test_docker_image() {
     local image=$1
     local container_name=$2
     local expected_distribution_type=$3
+    local expected_version=$4
 
-    if [ "$(docker ps --all --quiet --filter name="$container_name")" ]; then
-      echo "Removing existing '$container_name' container"
-      docker container rm --force "$container_name"
+    remove_container_if_exists "${container_name}"
+
+    check_distribution_type "${image}" "${expected_distribution_type}"
+    check_version "${image}" "${expected_version}"
+
+    test_map_read_write "${image}" "${container_name}"
+}
+
+function remove_container_if_exists() {
+    local container_name=$1
+
+    local containers
+    containers=$(docker ps --all --quiet --filter name="${container_name}")
+
+    if [[ -n ${containers} ]]; then
+      echo "Removing existing '${container_name}' container"
+      docker container rm --force "${container_name}"
     fi
+}
 
-    echo "Checking if $image is EE"
-    if docker run --rm $image bash -c 'compgen -G lib/*enterprise*'; then
+function check_distribution_type() {
+    local image=$1
+    local expected_distribution_type=$2
+
+    echo "Checking if ${image} is EE"
+    if docker run --rm "${image}" bash -c 'compgen -G lib/*enterprise*'; then
       echo "EE contents identified"
       distribution_type="ee"
     else
@@ -22,28 +41,46 @@ function test_docker_image() {
       distribution_type="oss"
     fi
 
-    if [[ "$distribution_type" != "$expected_distribution_type" ]]; then
-      echo "Distribution was $distribution_type, not $expected_distribution_type as expected"
+    if [[ "${distribution_type}" != "${expected_distribution_type}" ]]; then
+      echo "Distribution was ${distribution_type}, not ${expected_distribution_type} as expected"
       exit 1
     fi
+}
 
-    echo "Starting container '$container_name' from image '$image'"
-    docker run -it --name "$container_name" -e HZ_LICENSEKEY -e HZ_INSTANCETRACKING_FILENAME -d -p5701:5701 "$image"
+function check_version() {
+    local image=$1
+    local expected_version=$2
+
+    echo "Checking ${image} version"
+    local version
+    version=$(docker run --rm hazelcast/hazelcast-enterprise:latest-snapshot bin/hz-cli --version | awk '/Hazelcast/ {print $2}')
+    if [[ "${version}" != "${expected_version}" ]]; then
+      echo "${image} was ${version}, not ${expected_version} as expected"
+      exit 1
+    fi
+}
+
+function test_map_read_write() {
+    local image=$1
+    local expected_distribution_type=$2
+
+    echo "Starting container '${container_name}' from image '${image}'"
+    docker run -it --name "${container_name}" -e HZ_LICENSEKEY -e HZ_INSTANCETRACKING_FILENAME -d -p5701:5701 "${image}"
     local key="some-key"
     local expected="some-value"
-    echo "Putting value '$expected' for key '$key'"
-    while ! clc --timeout 5s map set -n some-map $key $expected --log.path stderr
+    echo "Putting value '${expected}' for key '${key}'"
+    while ! clc --timeout 5s map set -n some-map "${key}" "${expected}" --log.path stderr
     do
       echo "Retrying..."
       sleep 3
     done
-    echo "Getting value for key '$key'"
+    echo "Getting value for key '${key}'"
     local actual
-    actual=$(clc map get --format delimited -n some-map $key --log.path stderr)
-    echo "Stopping container $container_name}"
-    docker stop "$container_name"
+    actual=$(clc map get --format delimited -n some-map "${key}" --log.path stderr)
+    echo "Stopping container ${container_name}"
+    docker stop "${container_name}"
 
-    if [ "$expected" != "$actual" ]; then
+    if [[ "${expected}" != "${actual}" ]]; then
         echo "Expected to read '${expected}' but got '${actual}'"
         exit 1;
     fi
@@ -55,9 +92,9 @@ function install_clc() {
       echo "Retrying clc installation..."
       sleep 3
     done
-  export PATH=$PATH:$HOME/.hazelcast/bin
+  export PATH=${PATH}:${HOME}/.hazelcast/bin
   clc config add default cluster.name=dev cluster.address=localhost
 }
 
-install_clc
+# install_clc
 test_docker_image "$@"
--- a/.github/scripts/simple-smoke-test.sh
+++ b/.github/scripts/simple-smoke-test.sh
@@ -32,7 +32,7 @@ function check_distribution_type() {
     local image=$1
     local expected_distribution_type=$2
 
-    echo "Checking if ${image} is EE"
+    echo "Checking ${image} distribution type"
     if docker run --rm "${image}" bash -c 'compgen -G lib/*enterprise*'; then
       echo "EE contents identified"
       distribution_type="ee"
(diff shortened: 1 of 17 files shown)


.github/scripts/simple-smoke-test.sh
```diff
@@ -1,49 +1,86 @@
...
 
-    if [ "$(docker ps --all --quiet --filter name="$container_name")" ]; then
-      echo "Removing existing '$container_name' container"
-      docker container rm --force "$container_name"
+    remove_container_if_exists "${container_name}"
+
+    check_distribution_type "${image}" "${expected_distribution_type}"
+    check_version "${image}" "${expected_version}"
+
+    test_map_read_write "${image}" "${container_name}"
```
ldziedziul, line 16: Nice cleanup 💪

.github/scripts/simple-smoke-test.sh
```diff
@@ -1,49 +1,86 @@
...
+}
+
+function check_distribution_type() {
+    local image=$1
+    local expected_distribution_type=$2
 
-    echo "Checking if $image is EE"
-    if docker run --rm $image bash -c 'compgen -G lib/*enterprise*'; then
+    echo "Checking if ${image} is EE"
+    if docker run --rm "${image}" bash -c 'compgen -G lib/*enterprise*'; then
```
ldziedziul, line 36: Why without brackets? 

```suggestion
    if [ docker run --rm "${image}" bash -c 'compgen -G lib/*enterprise*' ]; then
```

.github/scripts/simple-smoke-test.sh
```diff
@@ -1,49 +1,86 @@
...
     fi
+}
+
+function check_distribution_type() {
+    local image=$1
+    local expected_distribution_type=$2
 
-    echo "Checking if $image is EE"
-    if docker run --rm $image bash -c 'compgen -G lib/*enterprise*'; then
+    echo "Checking if ${image} is EE"
```
ldziedziul, line 35: This is confusing, maybe something like: 
```suggestion
    echo "Checking type of the installed distribution in ${image}"
```

.github/scripts/simple-smoke-test.sh
```diff
@@ -1,49 +1,86 @@
...
       distribution_type="ee"
     else
       echo "No EE contents identified - assuming OSS"
       distribution_type="oss"
     fi
 
-    if [[ "$distribution_type" != "$expected_distribution_type" ]]; then
-      echo "Distribution was $distribution_type, not $expected_distribution_type as expected"
+    if [[ "${distribution_type}" != "${expected_distribution_type}" ]]; then
+      echo "Distribution was ${distribution_type}, not ${expected_distribution_type} as expected"
```
ldziedziul, line 45:
```suggestion
      echo "Image ${image} should contain ${expected_distribution_type} distribution but ${distribution_type} was detected"
```

.github/scripts/simple-smoke-test.sh
```diff
@@ -1,49 +1,86 @@
...
+
+function check_version() {
+    local image=$1
+    local expected_version=$2
+
+    echo "Checking ${image} version"
+    local version
+    version=$(docker run --rm "${image}" bin/hz-cli --version | awk '/Hazelcast/ {print $2}')
+    if [[ "${version}" != "${expected_version}" ]]; then
+      echo "${image} was ${version}, not ${expected_version} as expected"
```
ldziedziul, line 58:
```suggestion
      echo "Image ${image} should have ${expected_version} version but got ${version}"
```

.github/scripts/simple-smoke-test.sh
JackPGreen: > Why without brackets?

Spellcheck didn't like it as it was. What's the difference?

E.G. https://www.shellcheck.net
```
#!/usr/bin/env bash

if [ docker run ]; then
   echo "EE contents identified"
fi
```

.github/scripts/simple-smoke-test.sh
JackPGreen: 2bce024376f2d52e5e1c948efd90609fb6c97cf4

.github/scripts/simple-smoke-test.sh
```diff
@@ -1,49 +1,86 @@
...
+      echo "Image ${image} should contain ${expected_distribution_type} distribution but ${distribution_type} was detected"
       exit 1
     fi
+}
+
+function check_version() {
+    local image=$1
+    local expected_version=$2
+
+    echo "Checking ${image} version"
```
ldziedziul, line 54: I think it would be good to print what version was detected. 
image

JackPGreen: c037fde1ea66d9b57c9feaa45b38f1819d2605d5

ldziedziul (approved):

nishaatr:
Fix PR desc Spellcheck --> Shellcheck (may be add link)

JackPGreen:
> Fix PR desc Spellcheck --> Shellcheck (may be add link)

I genuinely thought it was Spellcheck 🤦

.github/scripts/simple-smoke-test.sh
```diff
@@ -1,49 +1,88 @@
...
+    test_map_read_write "${image}" "${container_name}"
+}
+
+function remove_container_if_exists() {
+    local container_name=$1
+
+    local containers
+    containers=$(docker ps --all --quiet --filter name="${container_name}")
+
+    if [[ -n ${containers} ]]; then
```
nishaatr, line 25: any chance could be more than one running container?

nishaatr, line 25: use "" ?

.github/scripts/simple-smoke-test.sh
```diff
@@ -1,49 +1,88 @@
...
       distribution_type="ee"
     else
       echo "No EE contents identified - assuming OSS"
       distribution_type="oss"
     fi
 
-    if [[ "$distribution_type" != "$expected_distribution_type" ]]; then
-      echo "Distribution was $distribution_type, not $expected_distribution_type as expected"
+    if [[ "${distribution_type}" != "${expected_distribution_type}" ]]; then
+      echo "Image ${image} should contain ${expected_distribution_type} distribution but ${distribution_type} was detected"
```
nishaatr, line 45: worth prepending with "ERROR: ...." so its clearer

.github/scripts/simple-smoke-test.sh
```diff
@@ -1,49 +1,88 @@
...
       exit 1
     fi
+}
+
+function check_version() {
```
nishaatr, line 50: may be rename to `check_hz_version()` or `check_image_hz_version()`

.github/scripts/simple-smoke-test.sh
```diff
@@ -1,49 +1,88 @@
...
+    local image=$1
+    local expected_version=$2
+
+    echo "Checking ${image} version"
+    local version
+    version=$(docker run --rm "${image}" bin/hz-cli --version | awk '/Hazelcast/ {print $2}')
+    if [[ "${version}" == "${expected_version}" ]]; then
+      echo "${image} version identified as ${version}"
+    else
+      echo "${image} version was ${version}, not ${expected_version} as expected"
```
nishaatr, line 60: "ERROR: ...."

.github/scripts/simple-smoke-test.sh
```diff
@@ -1,49 +1,88 @@
...
+    echo "Starting container '${container_name}' from image '${image}'"
+    docker run -it --name "${container_name}" -e HZ_LICENSEKEY -e HZ_INSTANCETRACKING_FILENAME -d -p5701:5701 "${image}"
     local key="some-key"
     local expected="some-value"
-    echo "Putting value '$expected' for key '$key'"
-    while ! clc --timeout 5s map set -n some-map $key $expected --log.path stderr
+    echo "Putting value '${expected}' for key '${key}'"
+    while ! clc --timeout 5s map set -n some-map "${key}" "${expected}" --log.path stderr
     do
       echo "Retrying..."
```
nishaatr, line 76: why retries?
It should work the first time otherwise there is issue I would guess

.github/scripts/simple-smoke-test.sh
```diff
@@ -1,49 +1,88 @@
...
       sleep 3
```
nishaatr, line 77: If it fails then we should bail out I think?
don't rely on below get check. it will be clearer where the issue lies IMO

.github/scripts/simple-smoke-test.sh
```diff
@@ -1,49 +1,88 @@
...
     done
-    echo "Getting value for key '$key'"
+    echo "Getting value for key '${key}'"
     local actual
-    actual=$(clc map get --format delimited -n some-map $key --log.path stderr)
-    echo "Stopping container $container_name}"
-    docker stop "$container_name"
+    actual=$(clc map get --format delimited -n some-map "${key}" --log.path stderr)
+    echo "Stopping container ${container_name}"
+    docker stop "${container_name}"
```
nishaatr, line 83: What if this fails?
if not a problem then any need to stop as it will die eventually once smoke test finishes

.github/scripts/simple-smoke-test.sh
```diff
@@ -1,49 +1,88 @@
...
 function test_docker_image() {
     local image=$1
     local container_name=$2
     local expected_distribution_type=$3
+    local expected_version=$4
 
-    if [ "$(docker ps --all --quiet --filter name="$container_name")" ]; then
-      echo "Removing existing '$container_name' container"
-      docker container rm --force "$container_name"
+    remove_container_if_exists "${container_name}"
```
nishaatr, line 11: pls add comment as to why we need to remove
This will run in its own docker/agent machine so I would assume env is fresh. I am sure I am missing something so comment will be good

.github/scripts/simple-smoke-test.sh
```diff
@@ -1,49 +1,88 @@
...
       echo "Retrying..."
       sleep 3
     done
-    echo "Getting value for key '$key'"
+    echo "Getting value for key '${key}'"
     local actual
-    actual=$(clc map get --format delimited -n some-map $key --log.path stderr)
-    echo "Stopping container $container_name}"
-    docker stop "$container_name"
+    actual=$(clc map get --format delimited -n some-map "${key}" --log.path stderr)
```
nishaatr, line 81: The command itself could fail for whatever reason like HZ is not responding
I think we should distinguish that vs no value found

.github/scripts/simple-smoke-test.sh
```diff
@@ -55,7 +94,7 @@ function install_clc() {
       echo "Retrying clc installation..."
       sleep 3
     done
-  export PATH=$PATH:$HOME/.hazelcast/bin
+  export PATH=${PATH}:${HOME}/.hazelcast/bin
```
nishaatr, line 97: Nit - normally better to add `${HOME}/.hazelcast/bin` at the beginning

.github/scripts/simple-smoke-test.sh
```diff
@@ -55,7 +94,7 @@ function install_clc() {
...
   clc config add default cluster.name=dev cluster.address=localhost
```
nishaatr, line 98: What is the reason for this?
Add comment pls as feels out of place otherwise

.github/scripts/simple-smoke-test.sh
```diff
@@ -1,49 +1,88 @@
...
+    check_version "${image}" "${expected_version}"
+
+    test_map_read_write "${image}" "${container_name}"
+}
+
+function remove_container_if_exists() {
+    local container_name=$1
+
+    local containers
+    containers=$(docker ps --all --quiet --filter name="${container_name}")
```
nishaatr, line 23: check cmd error first?

.github/scripts/simple-smoke-test.sh
```diff
@@ -1,49 +1,88 @@
...
+
+    if [[ -n ${containers} ]]; then
+      echo "Removing existing '${container_name}' container"
+      docker container rm --force "${container_name}"
```
nishaatr, line 27: check for error?

nishaatr (changes requested):

.github/scripts/simple-smoke-test.sh
JackPGreen: > any chance could be more than one running container?

Not sure what you mean?
This is existing functionality.

.github/scripts/simple-smoke-test.sh
JackPGreen: > What is the reason for this? Add comment pls as feels out of place otherwise

What do you mean?
My change or the modifications to `PATH`?

.github/scripts/simple-smoke-test.sh
JackPGreen: > use "" ?

Sorry I don't get what you mean?

.github/scripts/simple-smoke-test.sh
JackPGreen: 96285fe27e33b1cecdb9a381a74befa5a54a7d53

.github/scripts/simple-smoke-test.sh
JackPGreen: > why retries?

https://github.com/hazelcast/hazelcast-docker/pull/757

.github/scripts/simple-smoke-test.sh
JackPGreen: > check cmd error first?

Not sure what you mean?
The script has `set -o errexit` so I think were safe.

.github/scripts/simple-smoke-test.sh
JackPGreen: > check for error?

The script has `set -o errexit` so I think were safe.

.github/scripts/simple-smoke-test.sh
JackPGreen: > pls add comment as to why we need to remove
This will run in its own docker/agent machine so I would assume env is fresh. I am sure I am missing something so comment will be good

This is existing functionality. I'm guessing it's from local debugging.

.github/scripts/simple-smoke-test.sh
JackPGreen: 41bfd4c8a48fb71df7a2e037bfbd58ebe397eb22

.github/scripts/simple-smoke-test.sh
JackPGreen: > If it fails then we should bail out I think?
don't rely on below get check. it will be clearer where the issue lies IMO

Are you saying we shouldn't poll/retry? Or something else?

.github/scripts/simple-smoke-test.sh
JackPGreen: > What if this fails? if not a problem then any need to stop as it will die eventually once smoke test finishes

The script has `set -o errexit` so I think were safe.

.github/scripts/simple-smoke-test.sh
JackPGreen: > The command itself could fail for whatever reason like HZ is not responding I think we should distinguish that vs no value found

I _think_ it'll work the same - whether the command prints `0`, `1` or `Failed to connect to Hazelcast cluster!`, we'll check that against `expected` and error & print if different.

.github/scripts/simple-smoke-test.sh
nishaatr: ah yes. good point
may be test it just to be sure?
I will resolve nonethelss

.github/scripts/simple-smoke-test.sh
nishaatr: ah yes. good point

.github/scripts/simple-smoke-test.sh
nishaatr: True its current functionality but i was just thinking could the command return more than one container running.
But now I am thinking can two containers of the name be running at the same time?

JackPGreen: > True its current functionality but i was just thinking could the command return more than one container running. But now I am thinking can two containers of the name be running at the same time?

https://stackoverflow.com/questions/66591150/use-multiple-dockers-with-same-name#:~:text=You%20cannot%20have%20multiple%20containers,Docs%20regarding%20my%20below%20explanation.

JackPGreen: 2dae20b81fea906d8d2ad150e1f4a063338d7863

nishaatr (changes requested):

.github/scripts/simple-smoke-test.sh
JackPGreen: Ah ok I see what you mean.

`clc config add default cluster.name=dev cluster` is existing functionality, I'm not sure why it's there. It was added in 65f4b3fbe39a5ce6d978ff68803dc4d67090efd9 if that helps.

.github/scripts/simple-smoke-test.sh
JackPGreen: > Understood but I am saying don't rely on `expected` later on if the issue is with cluster/docker then we should error out so its clear when debugging I think we should not conflate the getting and then the checking Thats how I normally do it

Sorry, I don't understand?

.github/scripts/simple-smoke-test.sh
nishaatr: thanks for confirming. resolving

.github/scripts/simple-smoke-test.sh
nishaatr: I am saying we should check for error after running `clc` and bail out if error rather then relying on the later steps when checking for key/value

JackPGreen: > I am saying we should check for error after running `clc` and bail out if error rather then relying on the later steps when checking for key/value

What error would you be looking to catch?

`install_clc` _should_ fail if the installation fails so beyond that we can assume it's working.

This is existing functionality btw.

nishaatr: any error from `clc` tbh
I think using `$(clc...)` captures the error and I am thinking script will not fail unless you check the error
May be I am wrong

ldziedziul: > any error from `clc` tbh I think using `$(clc...)` captures the error and I am thinking script will not fail unless you check the error May be I am wrong

but `actual` content will be wrong and test will fail

There's also `set -E` flag to catch subshell error if I'm not mistaken, see https://stackoverflow.com/a/72103511

.github/scripts/simple-smoke-test.sh
nishaatr: its minor so resolving

.github/scripts/simple-smoke-test.sh
nishaatr: yes it will fail as `actual` content will be wrong
But normal practice is to handle errors right where it happens
If `set -E` works then thats ok too
Otherwise, don't want to drag so happy as it is!

JackPGreen: > yes it will fail as `actual` content will be wrong But normal practice is to handle errors right where it happens If `set -E` works then thats ok too Otherwise, don't want to drag so happy as it is!

I've tested it with a simple bash script:
```
#!/usr/bin/env bash

set -o errexit -o nounset -o pipefail

actual=$(clc map get --format delimited -n some-map this-map-doesnt-exist --log.path stderr)
echo finished
```

And it sets the parent errorlevel when the script exits

nishaatr: thanks for confirming. will resolve

JackPGreen: > thanks for confirming. will resolve

Thanks - I think that was the last outstanding issue from you?

.github/scripts/simple-smoke-test.sh
nishaatr: we should retry but should bail out after some retry count?

.github/scripts/simple-smoke-test.sh
nishaatr: Last one https://github.com/hazelcast/hazelcast-docker/pull/776/files#r1668945722
Just commented

.github/scripts/simple-smoke-test.sh
JackPGreen: > we should retry but should bail out after some retry count?

I don’t disagree but this isn’t new for this PR. The action has a timeout anyway (albeit 6 hours I think…)

nishaatr: ok. i will resolve

PR #779, opened 2024-07-08
Use centralised `assert.sh` script [DI-171]

JackPGreen:
The `assert.sh` script was [sourced externally](https://github.com/torokmark/assert.sh) but a local copy was included in the repository.

This change removes the local copy, instead dynamically fetching from a central [Hazelcast fork](https://github.com/hazelcast/assert.sh).

Our local copy had some small changes, which have been lost, but the previous version has been migrated to a branch of the Hazelcast fork to allow [a diff of the script](https://github.com/hazelcast/assert.sh/compare/main...hazelcast-docker).

Specifically, the changes are only to improve formatting of assertion messages, not the assertions themselves.

Fixes: [DI-171](https://hazelcast.atlassian.net/browse/DI-171)

[DI-171]: https://hazelcast.atlassian.net/browse/DI-171?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ldziedziul (changes requested):
Have you actually test if it works? 

Your branch: 
image

`master` branch:
image

JackPGreen:
> Have you actually test if it works?
> 
> Your branch: image
> 
> `master` branch: image

Yes, I've tested it - one of the changes we've lost is positive confirmation when the tests pass.

If you deliberately introduce an assertion failure:
image

ldziedziul:
> Yes, I've tested it - one of the changes we've lost is positive confirmation when the tests pass.

Thanks. Too bad... it doesn't build confidence for tests. Either we add a PR for the upstream or maybe it's another argument for [bashunit](http://bashunit.typeddevs.com)

.github/scripts/build.functions_tests.sh
```diff
@@ -4,7 +4,9 @@ set -eu
 
 SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"
 
-. "$SCRIPT_DIR"/assert.sh/assert.sh
+assert_script_content=$(curl --silent https://raw.githubusercontent.com/hazelcast/assert.sh/main/assert.sh)
+# shellcheck source=/dev/null
+. <(echo "${assert_script_content}")
```
nishaatr, line 9: Can you explain what these two lines do?
```
# shellcheck source=/dev/null
. <(echo "${assert_script_content}")
```
Also, what happens if `curl` fails?

nishaatr (changes requested):

.github/scripts/build.functions_tests.sh
JackPGreen: Currently we source (`.`) the content of another script on the local filesystem into the environment.

This change:
- downloads the latest copy of the script from our fork, saving the content to a variable
- feeds that into the source mechanism, as if the variable content was actually a file ([more info](https://superuser.com/a/1060002/1220887))

JackPGreen:
> Why this change [993c7cb](https://github.com/hazelcast/hazelcast-docker/commit/993c7cb8f0f2cf4ed37a199f7fa9daa779d26641) ?

I was testing the [changes discussed here](https://github.com/hazelcast/hazelcast-docker/pull/779#discussion_r1668846294) but focussed on getting it working and forgot to give the variable a sensible name until after I looked back at the PR.

.github/scripts/build.functions_tests.sh
JackPGreen: > Also, what happens if curl fails?

Sorry, missed that. 

_Should_ fail, because the script has `set -e`.
That's the reason for splitting out download from source mechanism, to ensure non-zero exit codes are captured.

I've just tested it with a wrong URL, it does give a non-zero error code from the `test_scripts.sh` script - which means the GitHub actions using it will fail.
No error message, though.

JackPGreen: > Interesting but little confusing must say. Can we add a comment OR simply create file in the same location as before (pass location to curl)?

I'm not saying it's super readable as it is now (mostly because `bash`), but I'm not sure what comment would help comprehension. Did you have any ideas?

nishaatr: May be point to https://github.com/hazelcast/hazelcast-docker/pull/779#discussion_r1668846294?
Otherwise its ok as I understand it now!

JackPGreen: 43f6296c68e0f911d84a11854e7c823b2660e708

JackPGreen:
> > Yes, I've tested it - one of the changes we've lost is positive confirmation when the tests pass.
> 
> Thanks. Too bad... it doesn't build confidence for tests. Either we add a PR for the upstream or maybe it's another argument for [bashunit](http://bashunit.typeddevs.com)

Having a further look, I think our usage of assertions are wrong.

In their example:
https://github.com/torokmark/assert.sh/blob/553bded1ba57894499ff6a6b6fd53c62e99d8b40/test_assert.sh#L398-403

The return code of the function is checked after _each_ assertion, and it's the callers responsibility to print a pass message.
Our test methods print a failure message only, relying on the (modified) framework to print the pass message.

You could _probably_ replicate this failure by refactoring all the assertions' ternary's to support.

nishaatr (approved):

JackPGreen:
@ldziedziul are you able to take another look at this, please?

JackPGreen:
> @ldziedziul are you able to take another look at this, please?

👀

ldziedziul (changes requested):
> Yes, I've tested it - one of the changes we've lost is positive confirmation when the tests pass.

@JackPGreen could we patch our `assert.sh` fork to re-add it? I find it very useful to see that tests were actuall executed

ldziedziul:
Why stick to the upstream at any cost? 

We have a fork to modify it if needed. There's no movement with the PR you've submitted to the upstream.

> - but we miss out on any upstream fixes
There's no active development from what I see in the upstream git history

As for other frameworks reporting test executions:
- JUnit reports test execution to files or to stdout
- we wrap assert.sh with additional logic (related to `TESTS_RESULT` variable), I'm concerned that could be broken in future
- when developing the tests I don't see if I actually run any tests

> We fully develop the `log_success` feature for all assertions and raise that as an upstream PR

I would say simply put `log_success` to all assertions in our fork (`main` branch) and use it in our projects

JackPGreen:
> Why stick to the upstream at any cost?
> 
> We have a fork to modify it if needed. 

Personal preference - I don't want to be a maintainer of a bash unit testing library.

> > * but we miss out on any upstream fixes
> There's no movement with the PR you've submitted to the upstream.

:(

> > We fully develop the `log_success` feature for all assertions and raise that as an upstream PR
> 
> I would say simply put `log_success` to all assertions in our fork (`main` branch) and use it in our projects

Look like a vote for (3)

@nishaatr any opinions?

nishaatr:
> @nishaatr any opinions?

I have been thinking about this and I am on the fence also. You both make a good case but I am leaning towards @ldziedziul for these reasons

1. We already have the fork
2. I thought we forked so we can apply and use our changes; otherwise, there was no point to fork
3. Doesn't look like the `assert.h` project receives frequent changes
4. As @ldziedziul has said, will be useful to see what successful tests were run - this happens in Jenkins as well
5. Also, the lines `======= Tests .... =======` having nothing under it looks odd and serves no purpose. Might as well don't emit or say something like `All tests passed` which will require a change
6. 
image
7. Our changes are small so easy to add

So I think lets backport changes to our fork and use that but also keep the PR open to upstream. Not sure we should do 4).
I prefer to use forked `main` also. Then we wait for upstream PR to be merged and then we can get rid of our fork or hopefully by that time we will know more about BashUnit

JackPGreen:
> I would say simply put `log_success` to all assertions in our fork (`main` branch) and use it in our projects

> I have been thinking about this and I am on the fence also. You both make a good case but I am leaning towards @ldziedziul for these reasons

Ok - cool, clear vote for assertion messages - so we'll do that.

I've had another idea how to achieve it - looking at [their documentation](https://github.com/torokmark/assert.sh?tab=readme-ov-file#how-to-write-tests), we _should_ be calling `log_success` ourselves after a successful assertion.

I've refactored the tests to do that.

1466d5015e9ad6e8a2a4b6b1b720a617f3f2cb20

That also means no upstream changes, or to our fork etc.

And the output now is as we want:
image

If your happy with that I'll roll that out across the other repositories.

nishaatr:
> I've had another idea how to achieve it - looking at [their documentation](https://github.com/torokmark/assert.sh?tab=readme-ov-file#how-to-write-tests), we _should_ be calling `log_success` ourselves after a successful assertion.

Thanks @JackPGreen 
So do we need the fork (apart for the upstream PR)?

JackPGreen:
> Thanks @JackPGreen So do we need the fork (apart for the upstream PR)?

No - we forked for security, not functionality. We aren't even using the PR branch here.

nishaatr (approved):

ldziedziul:
> That also means no upstream changes, or to our fork etc.

But means a little overhead in every test we create, nevertheless I can live with that

PR #801, opened 2024-09-12
Add support for latest-lts tags [DI-258][REL-343][5.5.1]

ldziedziul:
Adds support for latest-lts tags.

Whenever a `.github/latest_lts` file exists the build will push `latest-lts` tags and its variants (e.g. `latest-lts-jdk21`, `latest-lts-slim`, `latest-lts-slim-jdk21`, etc)

For manual runs you need to use `IS_LATEST_LTS` input instead.

PR #807, opened 2024-10-14
Add support for latest-lts tags [DI-258][REL-349][5.5.2]

ldziedziul:
Forwardport of #801 
Adds support for latest-lts tags.

Whenever a `.github/latest_lts` file exists the build will push
`latest-lts` tags and its variants (e.g. `latest-lts-jdk21`,
`latest-lts-slim`, `latest-lts-slim-jdk21`, etc)

For manual runs you need to use `IS_LATEST_LTS` input instead.


.github/latest_lts
```diff
@@ -0,0 +1 @@
+LTS
```
JackPGreen, line 1: Should we _also_ remove https://github.com/hazelcast/hazelcast-docker/blob/5.5.1/.github/latest_lts?

ldziedziul: Don't need it to as it won't be rebuilt after 5.5.2 release (we rebuild only the latest patch versions). There's a slight risk window but I've reworked this logic in the upcoming PR

JackPGreen (approved):

nishaatr (approved):

PR #811, opened 2024-10-23
Remove Hazelcast 4 RHEL support

JackPGreen:
Hazelcast `4.x` support in the workflow requires dynamically choosing credentials - but `4.x` is no longer supported, and this functionality complicates the workflow and can be removed.

Post-merge actions:
- [ ] delete corresponding GitHub secrets

PR #813, opened 2024-10-24
Run `tag_image_push` workflows during PR builder [DI-315]

JackPGreen:
The `tag_image_push` workflows are triggered _only_ with a release, which means that any changes are not able to be tested.

Using a `DRY_RUN` mode we can execute the workflow as a non-destructive sanity check, _without_ making any publicly observable changes.

Changes:
- Add execution of `tag_image_push` workflows (in `DRY_RUN` mode) to PR builder
- Extend `tag_image_push` workflow's `DRY_RUN` parameter by also skipping Dockerhub description/readme updates
- Support calling `tag_image_push` workflows from other workflows via `workflow_call` ([requires duplication of parameters](https://github.com/orgs/community/discussions/39357))
- update how we query `inputs` (`github.event.inputs` -> `inputs`) [to support `workflow_call` executions](https://github.blog/changelog/2022-06-10-github-actions-inputs-unified-across-manual-and-reusable-workflows/)
- give log artifacts more unique names now were running more jobs at once to avoid collisions

Fixes: [DI-315](https://hazelcast.atlassian.net/browse/DI-315)

Post-merge actions:
- [ ] backport

[DI-315]: https://hazelcast.atlassian.net/browse/DI-315?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -120,6 +120,73 @@ jobs:
   jdks:
     uses: ./.github/workflows/get-supported-jdks.yaml
 
+  build-pr-custom-jdk:
+    runs-on: ubuntu-latest
+    needs: [ jdks, prepare ]
+    name: Build with jdk-${{ matrix.jdk }}
+    strategy:
+      fail-fast: false
+      matrix:
+        jdk: ${{ fromJSON(needs.jdks.outputs.jdks) }}
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up Docker
+        uses: docker/setup-buildx-action@v3
+
+      - name: Build OSS image
+        run: |
+          DOCKER_PATH=hazelcast-oss
+          HZ_VERSION="${{ needs.prepare.outputs.HZ_VERSION_OSS }}"
+          
+          # duplicated block as GH doesn't support passing sensitive data between jobs
+          . .github/scripts/oss-build.functions.sh
+          export HZ_SNAPSHOT_INTERNAL_PASSWORD=${{ secrets.HZ_SNAPSHOT_INTERNAL_PASSWORD }}
+          export HZ_SNAPSHOT_INTERNAL_USERNAME=${{ secrets.HZ_SNAPSHOT_INTERNAL_USERNAME }}
+          HAZELCAST_OSS_ZIP_URL=$(get_hz_dist_zip "" "${HZ_VERSION}")
+          
+          curl --fail --silent --show-error --location "$HAZELCAST_OSS_ZIP_URL" --output $DOCKER_PATH/hazelcast-distribution.zip;
+          docker buildx build --load \
+          --build-arg JDK_VERSION=${{ matrix.jdk }} \
+          --build-arg HZ_VERSION=${HZ_VERSION} \
+          --tag hazelcast-oss:test \
+          ${DOCKER_PATH}
+      - name: Run smoke test against OSS image
+        timeout-minutes: 2
+        run: |
+          .github/scripts/simple-smoke-test.sh hazelcast-oss:test ${{ env.test_container_name_oss }} oss ${{ needs.prepare.outputs.HZ_VERSION_OSS }}
+
+      - name: Build Test EE image
+        run: |
+          DOCKER_PATH=hazelcast-enterprise
+          HZ_VERSION="${{ needs.prepare.outputs.HZ_VERSION_EE }}"
+          curl --fail --silent --show-error --location "${{ needs.prepare.outputs.HAZELCAST_EE_ZIP_URL }}" --output $DOCKER_PATH/hazelcast-enterprise-distribution.zip;
+          docker buildx build --load \
+          --build-arg JDK_VERSION=${{ matrix.jdk }} \
+          --build-arg HZ_VERSION=${HZ_VERSION} \
+          --tag hazelcast-ee:test \
+          ${DOCKER_PATH}
+      - name: Run smoke test against EE image
+        timeout-minutes: 2
+        run: |
+          export HZ_LICENSEKEY=${{ secrets.HZ_ENTERPRISE_LICENSE }}
+          .github/scripts/simple-smoke-test.sh hazelcast-ee:test ${{ env.test_container_name_ee }} ee ${{ needs.prepare.outputs.HZ_VERSION_EE }}
+      - name: Get docker logs
+        if: ${{ always() }}
+        run: |
+          docker logs ${{ env.test_container_name_oss }} > ${{ env.docker_log_file_oss }}
+          docker logs ${{ env.test_container_name_ee }} > ${{ env.docker_log_file_ee }}
+      - name: Store docker logs as artifact
+        if: ${{ always() }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: docker-logs-jdk${{ matrix.jdk }}
+          path: |
+            ${{ env.docker_log_file_oss }}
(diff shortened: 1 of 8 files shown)


ldziedziul:
> * Invert `DRY_RUN` check so that it _only_ runs if it's `false` (previously - ran always unless `true`) as a safety net for bad inputs (e.g. ``!)

It was done on purpose to avoid backward incompatibility here:
- https://github.com/hazelcast/hazelcast-docker/blob/79f6dede5bc866e09d204ec6210251be8d2129c3/.github/workflows/check-base-images.yml#L115-L116

If you change the default value of `DRY_RUN` then you have to add `DRY_RUN` flag in the lines above, but this won't work when rebuilding older versions without this flag. So you would have to backport `DRY_RUN` to older releases

JackPGreen:
Note that the [RHEL push tests are failing](https://github.com/hazelcast/hazelcast-docker/actions/runs/11507503356/job/32033821698), which is a good thing.

It's flagged that our _current_ implementation only supports Hazelcast `v5`, not `v6` - but it's better to spot this now, than after release.

This is already addressed in https://github.com/hazelcast/hazelcast-docker/pull/811.

JackPGreen:
> > * Invert `DRY_RUN` check so that it _only_ runs if it's `false` (previously - ran always unless `true`) as a safety net for bad inputs (e.g. ``!)
> 
> It was done on purpose to avoid backward incompatibility here:
> 
> * https://github.com/hazelcast/hazelcast-docker/blob/79f6dede5bc866e09d204ec6210251be8d2129c3/.github/workflows/check-base-images.yml#L115-L116
> 
> If you change the default value of `DRY_RUN` then you have to add `DRY_RUN` flag in the lines above, but this won't work when rebuilding older versions without this flag. So you would have to backport `DRY_RUN` to older releases

👍

f1ca0b4f4f0ce0feb57e1849f254baafb273f1ee

nishaatr:
Is this in relation to discussion we had to reduce Docker image testing during release? 
https://hazelcast.atlassian.net/wiki/spaces/EN/pages/4681302104/Create+release+5.x#Docker-release-tag-test

BTW...JIRA needs parent, SP and description

nishaatr (changes requested):

.github/workflows/build-pr.yml
```diff
@@ -189,8 +194,28 @@ jobs:
...
           path: |
             ${{ env.docker_log_file_oss }}
             ${{ env.docker_log_file_ee }}
 
+  test-push:
+    name: Test pushing image (dry run)
+    needs: [ prepare ]
+    uses: ./.github/workflows/tag_image_push.yml
+    with:
+      HZ_VERSION: ${{ needs.prepare.outputs.LAST_RELEASED_HZ_VERSION_OSS }}
```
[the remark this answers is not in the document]
JackPGreen: When you build a Hazelcast Docker image without specifying a distribution ZIP to use, it defaults to the latest released version - this is... ok in the tag workflows as they are just after we release, but here it's going to be expecting a `6.0.0-SNAPSHOT` image when it's actually going to build a `5.5.0` image in this job, and then the smoke test fails.

So we update the expectation of the assertion.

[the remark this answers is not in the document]
nishaatr: thanks

nishaatr (approved):

JackPGreen:
@ldziedziul are you able to take a look at this, please?

.github/workflows/build-pr.yml
```diff
@@ -189,8 +194,28 @@ jobs:
...
+
+  test-push-rhel:
+    name: Test pushing image RHEL (dry run)
+    needs: [ prepare ]
+    uses: ./.github/workflows/tag_image_push_rhel.yml
+    with:
+      HZ_VERSION: ${{ needs.prepare.outputs.LAST_RELEASED_HZ_VERSION_OSS }}
+      RELEASE_VERSION: ${{ needs.prepare.outputs.LAST_RELEASED_HZ_VERSION_OSS }}
+      DRY_RUN: true
+    secrets: inherit
```
ldziedziul, line 221: I'm concerned that we pass all the secrets needed to do the actual push to the prod registry. What if one creates a PR and modifies `tag_image_push.yml` that DRY_RUN is not working anymore? We would end up with pushing broken image to prod. 

WDYT about passing here secrets explicitly and only the ones that are used for read-only operations?

JackPGreen: > WDYT about passing here secrets explicitly and only the ones that are used for read-only operations?

I was trying to avoid a change to the `push` action necessitating a change to `build-pr` as well, but I can see your concern. 

However, if they are passed explicitly, how are they accessed in the workflow? AFAIK with the `with` syntax it passes them as environment variables, so the `${{ secrets.MY_CRED }}" would fail if `MY_CRED` was instead an environment variable?

ldziedziul: There's also `secrets` section. Check the first example https://docs.github.com/en/enterprise-cloud@latest/actions/sharing-automations/reusing-workflows#passing-inputs-and-secrets-to-a-reusable-workflow: 
```
jobs:
  call-workflow-passing-data:
    uses: octo-org/example-repo/.github/workflows/reusable-workflow.yml@main
    with:
      config-path: .github/labeler.yml
    secrets:
      personal_access_token: ${{ secrets.token }}
```

JackPGreen: I've tried in f1f1a3f60753fd6e1df7a94e7656d2d5af27d1cd

However the RHEL `push` needs the credentials for pushing, but _also_ the preflight etc stuff we actually want to test, so it's still `inherit`.

JackPGreen:
@ldziedziul are you able to take a look, please?

JackPGreen:
> @ldziedziul are you able to take a look, please?

just a reminder :)

.github/workflows/tag_image_push_rhel.yml
```diff
@@ -23,6 +23,33 @@ on:
...
+        default: 'false'
+        type: choice
+        options:
+          - 'false'
+          - 'true'
+  workflow_call:
+    inputs:
+      HZ_VERSION:
+        type: string
+        description: 'Version of Hazelcast to build the image for, e.g. 5.1.1, 5.0.1, 4.2.3'
```
ldziedziul, line 37: Taking https://hazelcast.atlassian.net/browse/DI-342 into account, we would end up with unpublishing HZ_VERSION while testing PRs. I would suggest either
-  skip running `tag_image_push_rhel.yml` during PR checks
-  modify it to not push anything to RHEL in dry-run and run only `simple-smoke-test.sh` instead
-  push the image with a temp tag like `pr-check-813`, but it depends on the approach we'll choose to fix permanently DI-342

JackPGreen: > Taking https://hazelcast.atlassian.net/browse/DI-342 into account, we would end up with unpublishing HZ_VERSION while testing PRs. I would suggest either
> 
> * skip running `tag_image_push_rhel.yml` during PR checks
> * modify it to not push anything to RHEL in dry-run and run only `simple-smoke-test.sh` instead
> * push the image with a temp tag like `pr-check-813`, but it depends on the approach we'll choose to fix permanently DI-342

Agreed.

> * skip running `tag_image_push_rhel.yml` during PR checks


https://github.com/hazelcast/hazelcast-docker/pull/813/commits/a6e86d0c75f913f472b2915f17efd16701e3234f

ldziedziul (approved):
LGTM

PR #816, opened 2024-10-31
Test JDK version as part of PR builder [DI-317]

JackPGreen:
Adds additional assertion to the PR builder smoke test that the JDK is the version expected for the image.

This is currently a _manual, post release_ check, moving earlier in the cycle to catch defects earlier.

Fixes: [DI-317](https://hazelcast.atlassian.net/browse/DI-317)

Post merge actions:
- [ ] backport
- [ ] _Discuss_ if we remove the manual, post-release check?

[DI-317]: https://hazelcast.atlassian.net/browse/DI-317?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/scripts/simple-smoke-test.sh
+++ b/.github/scripts/simple-smoke-test.sh
@@ -31,7 +31,6 @@ function stop_container() {
 
 function check_java_version() {
     local expected_major_version=$1
-    echo "Checking Java version - expected ${expected_major_version}"
     local actual_major_version
     actual_major_version=$(docker run --rm "${image}" sh -c 'java -version 2>&1 | head -n 1 | awk -F "\"" "{print \$2}" | awk -F "." "{print \$1}"')
 
--- a/.github/scripts/simple-smoke-test.sh
+++ b/.github/scripts/simple-smoke-test.sh
@@ -34,7 +34,9 @@ function check_java_version() {
     local actual_major_version
     actual_major_version=$(docker run --rm "${image}" sh -c 'java -version 2>&1 | head -n 1 | awk -F "\"" "{print \$2}" | awk -F "." "{print \$1}"')
 
-    if [[ "${expected_major_version}" != "${actual_major_version}" ]]; then
+    if [[ "${expected_major_version}" == "${actual_major_version}" ]]; then
+      echo "Expected Java version (${expected_distribution_type}) identified."
+    else
       echoerr "Expected Java version '${expected_major_version}' but got '${actual_major_version}'"
       exit 1;
     fi
--- a/.github/scripts/simple-smoke-test.sh
+++ b/.github/scripts/simple-smoke-test.sh
@@ -1,7 +1,6 @@
 #!/usr/bin/env bash
 
 set -o errexit
-set -x
 
 # shellcheck source=../.github/scripts/abstract-simple-smoke-test.sh
 . .github/scripts/abstract-simple-smoke-test.sh
--- a/.github/scripts/simple-smoke-test.sh
+++ b/.github/scripts/simple-smoke-test.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
 set -o errexit
+set -x
 
 # shellcheck source=../.github/scripts/abstract-simple-smoke-test.sh
 . .github/scripts/abstract-simple-smoke-test.sh
@@ -29,6 +30,18 @@ function stop_container() {
     docker stop "${container_name}"
 }
 
+function check_java_version() {
+    local expected_major_version=$1
+    echo "Checking Java version - expected ${expected_major_version}"
+    local actual_major_version
+    actual_major_version=$(docker run --rm "${image}" sh -c 'java -version 2>&1 | head -n 1 | awk -F "\"" "{print \$2}" | awk -F "." "{print \$1}"')
+
+    if [[ "${expected_major_version}" != "${actual_major_version}" ]]; then
+      echoerr "Expected Java version '${expected_major_version}' but got '${actual_major_version}'"
+      exit 1;
+    fi
+}
+
 function derive_expected_distribution_type() {
   local input_distribution_type=$1
 
@@ -50,6 +63,7 @@ image=$1
 container_name=$2
 input_distribution_type=$3
 expected_version=$4
+expected_java_major_version=$5
 
 
 remove_container_if_exists
@@ -59,3 +73,4 @@ trap stop_container EXIT
 
 expected_distribution_type=$(derive_expected_distribution_type "${input_distribution_type}")
 test_package "${expected_distribution_type}" "${expected_version}"
+check_java_version "${expected_java_major_version}"
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -82,8 +82,9 @@ jobs:
 
       - name: Run smoke test against OSS image
         timeout-minutes: 2
-        run: |
-          .github/scripts/simple-smoke-test.sh hazelcast-oss:test ${{ env.test_container_name_oss }} oss ${{ needs.prepare.outputs.HZ_VERSION_OSS }}
+        run: |          
+          . .github/scripts/docker.functions.sh
+          .github/scripts/simple-smoke-test.sh hazelcast-oss:test ${{ env.test_container_name_oss }} oss ${{ needs.prepare.outputs.HZ_VERSION_OSS }} "$(get_default_jdk hazelcast-oss)"
 
       - name: Build Test EE image
         run: |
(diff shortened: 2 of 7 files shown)


.github/scripts/simple-smoke-test.sh
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
 set -o errexit
+set -x
```
nishaatr, line 4: nit - can be move to above line

.github/scripts/simple-smoke-test.sh
```diff
@@ -29,6 +30,18 @@ function stop_container() {
     docker stop "${container_name}"
 }
 
+function check_java_version() {
+    local expected_major_version=$1
+    echo "Checking Java version - expected ${expected_major_version}"
```
nishaatr, line 35: nit - move `echo` after all `local` declarations

.github/scripts/simple-smoke-test.sh
```diff
@@ -29,6 +30,18 @@ function stop_container() {
...
+    local actual_major_version
+    actual_major_version=$(docker run --rm "${image}" sh -c 'java -version 2>&1 | head -n 1 | awk -F "\"" "{print \$2}" | awk -F "." "{print \$1}"')
+
```
nishaatr, line 38: echo `actual_major_version` for debugging purposes

.github/scripts/simple-smoke-test.sh
```diff
@@ -29,6 +30,18 @@ function stop_container() {
     docker stop "${container_name}"
 }
 
+function check_java_version() {
+    local expected_major_version=$1
+    echo "Checking Java version - expected ${expected_major_version}"
+    local actual_major_version
+    actual_major_version=$(docker run --rm "${image}" sh -c 'java -version 2>&1 | head -n 1 | awk -F "\"" "{print \$2}" | awk -F "." "{print \$1}"')
```
nishaatr, line 37: nit - no need for `sh -c`?
i think default is `sh`

nishaatr, line 37: pls add a comment giving example as to what line we are extracting from

.github/workflows/build-pr.yml
```diff
@@ -82,8 +82,9 @@ jobs:
 
       - name: Run smoke test against OSS image
         timeout-minutes: 2
-        run: |
-          .github/scripts/simple-smoke-test.sh hazelcast-oss:test ${{ env.test_container_name_oss }} oss ${{ needs.prepare.outputs.HZ_VERSION_OSS }}
+        run: |          
+          . .github/scripts/docker.functions.sh
+          .github/scripts/simple-smoke-test.sh hazelcast-oss:test ${{ env.test_container_name_oss }} oss ${{ needs.prepare.outputs.HZ_VERSION_OSS }} "$(get_default_jdk hazelcast-oss)"
```
nishaatr, line 87: We don't deploy OSS for PATCH
Just checking this is accounted for?

nishaatr (changes requested):
I don't see changes related to RHEL?

.github/scripts/simple-smoke-test.sh
JackPGreen: Left over from debugging, removed

276e1ae1f93567526fa180fe3329a4e1834630c3

JackPGreen:
> I don't see changes related to RHEL?

What RHEL changes are you expecting?

.github/workflows/build-pr.yml
JackPGreen: > We don't deploy OSS for PATCH Just checking this is accounted for?

This isn’t a release - in the PR, we build the images and push them to a (local) registry - for both OS & EE, always.

nishaatr: understood.

nishaatr:
> What RHEL changes are you expecting?

Add test when for RHEL build.
Or do we not build RHEL for PR?

.github/scripts/simple-smoke-test.sh
ldziedziul: why?

Usually, keeping variable closer to their usage makes the code easier to understand

.github/scripts/simple-smoke-test.sh
ldziedziul: It's echoed here: https://github.com/hazelcast/hazelcast-docker/pull/816/files#diff-ca3c00912d3244c223ca46144b4cb9ebb73b370d71128734a230d320d8478468R39

.github/scripts/simple-smoke-test.sh
```diff
@@ -29,6 +29,18 @@ function stop_container() {
...
```
ldziedziul, line 36: Against which distribution and java version have you tested the extraction of the major number?

.github/scripts/simple-smoke-test.sh
nishaatr: the standard is to declare vars first - at least thats what I do
Helps seeing easily what input vars are needed. 
In other langs this happens implicitly via function declaration
its little distracting otherwise.

.github/scripts/simple-smoke-test.sh
nishaatr: But thats if they don't match
I am saying echo regardless

.github/scripts/simple-smoke-test.sh
ldziedziul: I think this can help with debugging: 
`set -o errexit ${RUNNER_DEBUG:+-x}`

JackPGreen:
> > What RHEL changes are you expecting?
> 
> Add test when for RHEL build. Or do we not build RHEL for PR?

Short answer - no.

Long answer, _from what I understand_ we build Docker images which we deploy to Docker Hub and RHEL _seperately_.

This PR is about the scope of the testing of the _images_.

Testing of the _deployment_ is part of https://github.com/hazelcast/hazelcast-docker/pull/813

.github/scripts/simple-smoke-test.sh
JackPGreen: This is actually another debugging statement I'd left in... removed.

96388d8549078104635f9047ebe655adfe537fe4

.github/scripts/simple-smoke-test.sh
JackPGreen: > Against which distribution and java version have you tested the extraction of the major number?

Local JDK21 + whatever JDKs are being used when building the various Hazelcast Docker images.

I actually copied this from https://github.com/hazelcast/hazelcast-cpp-client/blob/4a027a85f131d733b44d0ffb32b74b989269e0ec/scripts/start-rc.sh#L121

.github/scripts/simple-smoke-test.sh
nishaatr: separate PR gets my vote
BASH novice question - what does `${RUNNER_DEBUG:+-x}` actually do?

JackPGreen: > separate PR gets my vote

https://github.com/hazelcast/hazelcast-docker/pull/817

> BASH novice question - what does `${RUNNER_DEBUG:+-x}` actually do?

It expands to `-x` if `RUNNER_DEBUG` is defined as a variable.

.github/scripts/simple-smoke-test.sh
JackPGreen: ce1db912e6d94ee1e2c2e378574a333b3e894ad4

.github/scripts/simple-smoke-test.sh
nishaatr: got it. neat. thanks

.github/scripts/simple-smoke-test.sh
nishaatr: ignore. I misunderstood

JackPGreen: > nit - no need for `sh -c`? i think default is `sh`

Were passing the entire `java -version 2>&1 | head -n 1 | awk -F "\"" "{print \$2}" | awk -F "." "{print \$1}"` command into the Docker image to execute.

If you run just java -version 2>&1` on the Docker image, and then pipe that into `head`, [you get a bizarre SIGPIPE failure](https://stackoverflow.com/a/19120674) - but only when running on the GitHub runner, not locally 

_ideally_ we shouldn't execute those commands in the container as were assuming it's got `head`, `awk` etc. But I wasted many hours on this.

nishaatr: Understood. thanks

nishaatr (approved):

.github/scripts/simple-smoke-test.sh
ldziedziul: In this case `actual_major_version` is not an input var so there's no need to move it up and extend its scope. 

> the standard is to declare vars first - at least thats what I do
There's no such standard, maybe in C. 

It's about readability, scope, and maintainability. Take a look at this discussion:
https://softwareengineering.stackexchange.com/questions/113262/why-declare-variables-close-to-where-they-are-used

ldziedziul (approved):

PR #823, opened 2024-11-15
Improve `publish-rhel` script [DI-342]

JackPGreen:
The action [failed](https://github.com/hazelcast/hazelcast-docker/actions/runs/11831672090/attempts/1) but without providing enough information to understand _why_.

Changes:
- Log the `get_image` function response (JSON) in case of failure
- Improve failure annotation reporting (as per https://github.com/hazelcast/hazelcast-docker/pull/821)
- Address all Shellcheck warnings in case related
   - [`Declare and assign separately to avoid masking return values`](https://www.shellcheck.net/wiki/SC2155)
   - [`Prefer putting braces around variable references even when not strictly required`](https://www.shellcheck.net/wiki/SC2250)
   - [`Quote the rhs of = in [[ ]] to prevent glob matchingQuote the rhs of = in [[ ]] to prevent glob matching`](https://www.shellcheck.net/wiki/SC2053)

_Partially addresses_ [DI-342](https://hazelcast.atlassian.net/browse/DI-342)

Post-merge actions:
- [ ] backport

### DIFF
--- a/.github/scripts/publish-rhel.sh
+++ b/.github/scripts/publish-rhel.sh
@@ -1,6 +1,9 @@
 #!/bin/bash
 
-set -euo pipefail ${RUNNER_DEBUG:+-x}
+set -o errexit -o nounset -o pipefail ${RUNNER_DEBUG:+-x}
+
+# shellcheck source=../.github/scripts/abstract-simple-smoke-test.sh
+. .github/scripts/abstract-simple-smoke-test.sh
 
 get_image()
 {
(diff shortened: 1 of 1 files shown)


nishaatr (approved):

ldziedziul (commented):
Have you tested it?

JackPGreen:
> Have you tested it?

I hadn't - because it's destructive I wanted to wait for a review.

As it's already had one approval, I've run it [here](https://github.com/hazelcast/hazelcast-docker/actions/runs/11930895196) ✅.

Unless you could think of a better way to test?

.github/scripts/publish-rhel.sh
```diff
@@ -1,6 +1,9 @@
 #!/bin/bash
 
-set -euo pipefail ${RUNNER_DEBUG:+-x}
+set -o errexit -o nounset -o pipefail ${RUNNER_DEBUG:+-x}
+
+# shellcheck source=../.github/scripts/abstract-simple-smoke-test.sh
+. .github/scripts/abstract-simple-smoke-test.sh
```
ldziedziul, line 6: Why do we need it?

JackPGreen: For the `echoerr` function.

ldziedziul: It's confusing that we have a test code in the production script. I would suggest either inline this function as it's very small or move it to another file `echo.functions.sh`/`log.functions.sh`?

JackPGreen: > It's confusing that we have a test code in the production script.

Good point.

> I would suggest either inline this function as it's very small or move it to another file `echo.functions.sh`/`log.functions.sh`?

For now I've duplicated it.
_Ideally_ it should be externalised - but [as `hazelcast-packaging` directly references the script without any other context](https://github.com/hazelcast/hazelcast-packaging/blob/9b5cd38369c7fdf883f6e9a58bb29c7e1a46b615/check-hazelcast-health.sh#L8), a larger change is required to support that.

PR #827, opened 2024-11-20
Fix `test_published_images.yml` [DI-316]

JackPGreen:
The action added in https://github.com/hazelcast/hazelcast-docker/pull/814 could not be properly tested via GitHub until merged and at that point additional defects were discovered.

[Example of now successful completion](https://github.com/hazelcast/hazelcast-docker/actions/runs/11943342729).

Post-merge actions:
- [ ] backport this + previous PRs to maintenance branches
- [ ] update [release documentation to reference](https://hazelcast.atlassian.net/wiki/spaces/EN/pages/4681302104/Create+release+5.x#Docker-release-tag-test)

Fixes: [DI-316](https://hazelcast.atlassian.net/browse/DI-316)

[DI-316]: https://hazelcast.atlassian.net/browse/DI-316?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/test_published_images.yml
+++ b/.github/workflows/test_published_images.yml
@@ -5,13 +5,13 @@ on:
     inputs:
       IMAGE_VERSION:
         required: true
-        description: A version, `5.4.1`, `latest` etc
+        description: The version/label of the image e.g. `5.4.1`, `latest` etc
       EXPECTED_HZ_VERSION:
-        description: A fully-qualified version, e.g. `5.4.1`
+        description: The expected Hazelcast version (fully-qualified), e.g. `5.4.1`
         required: true
       DISTRIBUTION_TYPE:
         required: true
-        description: The distribution(s) to test (case-sensitive)
+        description: The distribution(s) to test
         type: choice
         options:
           - oss
--- a/.github/workflows/test_published_images.yml
+++ b/.github/workflows/test_published_images.yml
@@ -52,8 +52,7 @@ jobs:
               exit 1
               ;;
           esac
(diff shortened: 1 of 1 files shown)


.github/workflows/test_published_images.yml
```diff
@@ -68,7 +68,7 @@ jobs:
         variant: 
           - ''
           - 'slim'
-        distribution_type: ${{ fromJson(needs.set-matrix.outputs.distribution-type-matrix) }}
+        distribution-type: ${{ fromJson(needs.set-matrix.outputs.distribution-type-matrix) }}
```
ldziedziul, line 71: why this rename?

JackPGreen:
> > The action added in #814 could not be properly tested via GitHub until merged
> 
> You could simply merge an empty (or minimalistic) workflow to the master, and then test the workflow on the branch

I wasn’t sure what the eventual inputs would be (which is good as we changed them right at the end) and was worried GitHub would only let me use the inputs on `master`.

.github/workflows/test_published_images.yml
JackPGreen: > why this rename?

Consistency between the matrix name and its source. Probably not required but neater. Can revert if problematic…

nishaatr (approved):

ldziedziul:
> I wasn’t sure what the eventual inputs would be (which is good as we changed them right at the end) and was worried GitHub would only let me use the inputs on `master`.

I'm pretty sure it will show input from the PR branch

ldziedziul (approved):

PR #838, opened 2024-11-27
Fix Actionlint warnings

JackPGreen:
[Actionlint](https://github.com/rhysd/actionlint) flagged warnings in our GitHub Actions:
- `.github/workflows/tag_image_push_rhel.yml:139:760: property "variant" is not defined in object type {jdk: any} [expression]`
   - The property is empty so implicitly evaluates to empty string, i.e. builds only "base" variant (not `slim`)
   - This is the behaviour we actually want, so made explicit

### DIFF
--- a/.github/workflows/tag_image_push.yml
+++ b/.github/workflows/tag_image_push.yml
@@ -17,6 +17,7 @@ on:
       RELEASE_TYPE:
         description: 'What should be built'
         required: true
+        default: 'EE'
         type: choice
         options:
           - ALL
--- a/.github/workflows/tag_image_push.yml
+++ b/.github/workflows/tag_image_push.yml
@@ -17,7 +17,6 @@ on:
       RELEASE_TYPE:
         description: 'What should be built'
         required: true
-        default: 'EE'
         type: choice
         options:
           - ALL
--- a/.github/workflows/tag_image_push_rhel.yml
+++ b/.github/workflows/tag_image_push_rhel.yml
@@ -158,7 +158,7 @@ jobs:
           docker buildx build --push \
             --build-arg HZ_VERSION=${{ env.HZ_VERSION }} \
             --build-arg JDK_VERSION=${{ matrix.jdk }} \
-            --build-arg HAZELCAST_ZIP_URL=$(get_hz_dist_zip "${{ matrix.variant }}" "${{ env.HZ_VERSION }}") \
+            --build-arg HAZELCAST_ZIP_URL=$(get_hz_dist_zip "" "${{ env.HZ_VERSION }}") \
             ${TAGS_ARG} \
             --platform=${PLATFORMS} $DOCKER_DIR
 


nishaatr (approved):
LGTM
Out of interest, when/where do we run [Actionlint](https://github.com/rhysd/actionlint)?

.github/workflows/tag_image_push.yml
```diff
@@ -17,7 +17,6 @@ on:
       RELEASE_TYPE:
         description: 'What should be built'
         required: true
-        default: 'EE'
```
ldziedziul, line 20: It's there for a reason, without it we would have `ALL` as the default value in the UI. Setting this value helps us avoiding OSS releases by accident

image

vs 

image

.github/workflows/tag_image_push_rhel.yml
```diff
@@ -158,7 +158,7 @@ jobs:
           docker buildx build --push \
             --build-arg HZ_VERSION=${{ env.HZ_VERSION }} \
             --build-arg JDK_VERSION=${{ matrix.jdk }} \
-            --build-arg HAZELCAST_ZIP_URL=$(get_hz_dist_zip "${{ matrix.variant }}" "${{ env.HZ_VERSION }}") \
+            --build-arg HAZELCAST_ZIP_URL=$(get_hz_dist_zip "" "${{ env.HZ_VERSION }}") \
```
ldziedziul, line 161: 👍

.github/workflows/tag_image_push.yml
nishaatr: What sets the order in UI?

JackPGreen:
> LGTM Out of interest, when/where do we run [Actionlint](https://github.com/rhysd/actionlint)?

We don't.

[In `hz-docs` we introduced a broken workflow](https://github.com/hazelcast/hz-docs/pull/1406). I thought [if the GitHub UI could flag this](https://github.com/hazelcast/hz-docs/actions/runs/12046739812/workflow), why couldn't we catch this _before_ it was merged via a GitHub Action linter or similar? So I tried a couple out to see if they'd help.
ActionLint is good _but_ it flags `shellcheck` warnings too - which is great, but we have lots - so difficult to introduce into an existing project.
Ultimately I didn't progress the idea any further.

.github/workflows/tag_image_push.yml
JackPGreen: > It's there for a reason, without it we would have `ALL` as the default value in the UI. Setting this value helps us avoiding OSS releases by accident

That makes sense, thanks.
If we were going to continue with ActionLint I'd document this nuance properly, but we aren't, so instead I'll just leave it alone.

4a78d0747be33ef05fbea54fe8dedeffd9113336

ldziedziul: > What sets the order in UI?

Order of items:
```yaml
        options:
          - ALL
          - OSS
          - EE
          - README
```

We could move `EE` to the top, but IMHO we would make the workflow fragile since adding a new value as a first option would break it. By making `EE` explicit we show our intention: this very option should be the default one. 

Also I've noticed that in https://jenkins.hazelcast.com/view/all/job/Hazelcast-packaging-master-snapshot-trigger/configure we don't specify release type which suggests that the default value is used there

nishaatr: Makes sense. good catch
Also remembered, for Release pipeline Jenkins build, EE is always checked

ldziedziul (approved):

PR #842, opened 2024-12-02
Centralise `echoerr` function

JackPGreen:
Currently is duplicated and should instead be centralised.

[Raised from PR feedback](https://github.com/hazelcast/hazelcast-docker/pull/823#discussion_r1865712579).

Requires https://github.com/hazelcast/hazelcast-packaging/pull/251.


.github/scripts/workflow-commands.functions.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+# https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions
+
+# Prints the given message to stderr
+function echoerr() {
```
nishaatr, line 6: rename `echo_err` or `echoErr` for better readability

.github/workflows/test_published_images.yml
```diff
@@ -37,6 +37,8 @@ jobs:
       - name: Parse input into matrix
         id: set-matrix
         run: |
+          . .github/scripts/workflow-commands.functions.sh
```
nishaatr, line 40: does this need `# shellcheck....`

.github/scripts/workflow-commands.functions.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
```
nishaatr, line 1: the name implies the file is only meant for GH workflow commands or any SH commands can be used?

nishaatr (changes requested):

.github/scripts/workflow-commands.functions.sh
JackPGreen: That would involve changing _every_ caller. I can do it… but is that valuable?

.github/workflows/test_published_images.yml
JackPGreen: Shellcheck doesn’t (natively) work on non-shell scripts (ie those embedded in other stuff). If you use something like action lint it would, but I don’t think anyone is so not sure of the value of adding here.

.github/scripts/workflow-commands.functions.sh
JackPGreen: It’s for GitHub API interface only.

ldziedziul: IMHO `logging.function.sh` would be more obvious

ldziedziul (approved):

.github/scripts/workflow-commands.functions.sh
nishaatr: lets leave it. Its just a nit.

.github/workflows/test_published_images.yml
nishaatr: Understood. But...

Here we have the spellcheck comment
**https://github.com/hazelcast/hazelcast-docker/blob/87cab87ce9ef71c7aff0047a6abf63c305de9cbb/.github/workflows/test_published_images.yml#L131**

And not here in the same file:
https://github.com/hazelcast/hazelcast-docker/blob/87cab87ce9ef71c7aff0047a6abf63c305de9cbb/.github/workflows/test_published_images.yml#L40

Both are non-shell

JackPGreen: Good spot. Personally I don’t _really_ mind which way it is, but I guess if I’m changing it it should be consistent. Any opinions either way?

nishaatr: I think lets add even if it is benign.

JackPGreen: https://github.com/hazelcast/hazelcast-docker/pull/842/commits/d435d7f22a4c0972bf3f6dd6db3f4afdbdab5432

nishaatr (approved):

PR #845, opened 2024-12-24
When `RHEL` publish fails, log reason (`test_results`)

JackPGreen:
While investigating an [action failure](https://github.com/hazelcast/hazelcast-docker/actions/runs/12407221911/attempts/1) of an `RHEL` publish failure, the logs indicated that the image had never published, but not _why_.

The publish response includes a link to an `test_results` result which can [provide more information](https://catalog.redhat.com/api/containers/docs/endpoints/RESTGetTestResultsByImageId.html).

Updated the script to log these results in the case of failure.

I've gone backwards and forwards on the implementation - _ideally_ the publish _request_ should be polled rather than the the presence of an image, but would require significant refactoring of the script which is difficult to test due to its destructive nature.
In fact, in isolation this change is pointless - we build from tags so the usage of this change will only gradually increase. But it helps to document _how_ we get investigate subsequent failures. Ideally once next investigated, the root cause can be identified and resolved.


.github/scripts/publish-rhel.sh
```diff
@@ -203,7 +206,21 @@ wait_for_container_publish()
...
             echoerr "Timeout! Publish could not be finished"
+            echoerr "Image Status:"
             echoerr "${IMAGE}"
+
+            # Add additional logging context if possible
+            echoerr "Test Results:"
+            # https://catalog.redhat.com/api/containers/docs/endpoints/RESTGetTestResultsById.html
+            get_image not_published "${RHEL_PROJECT_ID}" "${VERSION}" "${RHEL_API_KEY}" | jq -r '.data[]._links.test_results.href' | while read -r TEST_RESULTS_ENDPOINT; do
+                local TEST_RESULTS
+                TEST_RESULTS=$(curl --silent \
```
nishaatr, line 217: Tested and how?

.github/scripts/publish-rhel.sh
```diff
@@ -203,7 +206,21 @@ wait_for_container_publish()
 
         if [[ ${i} == "${NOF_RETRIES}" ]]; then
             echoerr "Timeout! Publish could not be finished"
+            echoerr "Image Status:"
             echoerr "${IMAGE}"
+
+            # Add additional logging context if possible
+            echoerr "Test Results:"
```
nishaatr, line 213: Didn't know tests are run during publish. Any ideas what kind of tests?
Thought tests would be run during scanning and not publishing

nishaatr (changes requested):

.github/scripts/publish-rhel.sh
JackPGreen: > Didn't know tests are run during publish. Any ideas what kind of tests? Thought tests would be run during scanning and not publishing

I think `tests` would be better thought of as `checks` - for example if you publish an image that references a URL, it checks the URL exists, else the check fails with `404`.

The full list of checks I couldn't say.

nishaatr: I think what I am trying to get at is that will `.data[]._links.test_results.href'` exists? 
Is it worth checking URL exists before making a `curl` call; otherwise hust skip?

JackPGreen: > I think what I am trying to get at is that will `.data[]._links.test_results.href'` exists? Is it worth checking URL exists before making a `curl` call; otherwise hust skip?

I think it will, but were in the error state already - the image hasn't published. So if we fail to get the explanation, that's ok too.

nishaatr (approved):

JackPGreen:

ldziedziul (approved):

PR #846, opened 2024-12-30
Fix `tag_image_push_rhel.yml` for `push` triggers [DI-375]

JackPGreen:
In https://github.com/hazelcast/hazelcast-docker/pull/844, the `HZ_VERSION` is computed using the `required` `HZ_VERSION` `input`.

However, this is only `required` for `workflow_dispatch` triggers - it can _also_ be triggered by `push`, where it's instead empty - causing an [action failure](https://github.com/hazelcast/hazelcast-docker/actions/runs/12521044660):
```
Run madhead/semver-utils@latest
Error: Input required and not supplied: version
```

The workflow already derived an appropriate `HZ_VERSION` environment variable by validating the one in the in the `input` - updated to use this instead as it _should_ have done in the first place.

Also simplified the above logic to avoid it overwriting `env.HZ_VERSION` with the same value pointlessly.

Fixes: [DI-375](https://hazelcast.atlassian.net/browse/DI-375)

[DI-375]: https://hazelcast.atlassian.net/browse/DI-375?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/tag_image_push_rhel.yml
```diff
@@ -48,22 +48,14 @@ jobs:
...
       - name: Set HZ version as environment variable
+        if: env.HZ_VERSION == ''
         run: |
-          if [ -z "${{ env.HZ_VERSION }}" ]; then
-             HZ_VERSION=${GITHUB_REF:11}
-          else
-             HZ_VERSION=${{ env.HZ_VERSION }}
-          fi
-          echo "HZ_VERSION=${HZ_VERSION}" >> $GITHUB_ENV
+          echo "HZ_VERSION=${GITHUB_REF:11}" >> $GITHUB_ENV
```
nishaatr, line 53: Just checking that this is extracting the version from the tag?

.github/workflows/tag_image_push_rhel.yml
```diff
@@ -87,7 +79,7 @@ jobs:
       - uses: madhead/semver-utils@latest
         id: version
         with:
-          version: ${{ inputs.HZ_VERSION }}
+          version: ${{ env.HZ_VERSION }}
```
nishaatr, line 82: Wondering if `lenient: false` should be set so fails in case version is not formatted properly (when running manually)?

nishaatr (changes requested):

.github/workflows/tag_image_push_rhel.yml
JackPGreen: From the `ref` - so could be branch (e.g. `master`).
It's existing functionality btw, just refactored.

.github/workflows/tag_image_push_rhel.yml
JackPGreen: > Wondering if `lenient: false` should be set so fails in case version is not formatted properly (when running manually)?

We don't need that elsewhere, so 🤞 it's all ok

nishaatr (approved):

JackPGreen:

ldziedziul (approved):

PR #847, opened 2024-12-30
Test `RHEL` v6 Hazelcast container [DI-391]

JackPGreen:
We need to way to test the credentials for the new `RHEL` v6 container introduced in [DI-319](https://hazelcast.atlassian.net/browse/DI-319) in a non-destructive way - in the same vein as https://github.com/hazelcast/hazelcast-docker/pull/813

Changes:
- Add `tag_image_push_rhel` to PR builder
- add a `DRY_RUN` mode that runs only the non-destructive elements, ensuring our credentials are correct
- ensure `DRY_RUN` is properly supported in the case of a `push` event where it's empty (overlooked in https://github.com/hazelcast/hazelcast-docker/pull/813)
- fixes ordering issue introduced in https://github.com/hazelcast/hazelcast-docker/pull/844 where `RHEL_PROJECT_ID` is referenced before it's generated

Fixes: [DI-391](https://hazelcast.atlassian.net/browse/DI-391)

[DI-391]: https://hazelcast.atlassian.net/browse/DI-391?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DI-319]: https://hazelcast.atlassian.net/browse/DI-319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nishaatr (changes requested):
Wondering if DRY_RUN is needed simply for testing creds?
I see why this time for testing RHEL v6 container (which is a rare event)
But generally, how often will the credentials change? If they do the build will fail and we can fix. 
The credentials can be tested locally (although I am aware we access AWS creds but thats a proven functionality). Saying this as now we have many IF checks for DRY_RUN and feels little excessive for just checking creds

That said, not too concerned as may be we can use DRY_RUN for other things in the future.

.github/scripts/docker.functions.sh
```diff
@@ -5,6 +5,11 @@ function get_default_jdk() {
   awk -F= '/^ARG JDK_VERSION=/{print $2}' "$DIR/Dockerfile" | tr -d '"'
 }
 
+function get_hz_version() {
```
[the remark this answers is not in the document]
JackPGreen: 852c80b39a649bfa4f4b26b47064c3116ab6400a

.github/workflows/tag_image_push_rhel.yml
```diff
@@ -24,6 +24,14 @@ on:
...
           - 'false'
           - 'true'
+      DRY_RUN:
+        description: 'Skip pushing the images to remote registry'
+        default: 'false'
+        type: choice
+        options:
+          - 'false'
+          - 'true'
+  pull_request:
```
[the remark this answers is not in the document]
JackPGreen: > why `pull_request:`?

To add this to the test suite that's run when you raise a PR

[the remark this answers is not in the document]
nishaatr: You mean run https://github.com/hazelcast/hazelcast-docker/blob/852c80b39a649bfa4f4b26b47064c3116ab6400a/.github/workflows/tag_image_push_rhel.yml#L186?

We don't do this for https://github.com/hazelcast/hazelcast-docker/blob/852c80b39a649bfa4f4b26b47064c3116ab6400a/.github/workflows/tag_image_push.yml (well I don't see `pull_request:`)

Also, DRY_RUN default is `false` so things will happen for PR too?

nishaatr:
> We aren't just testing credentials - some of the other steps are run too, e.g. deriving HZ_VERSION, checking the yml has valid syntax etc. It's not a great test, but it's something.

Sure but main bulk of functionality is now under DRY_RUN so in reality things might still fail during release

> The first time the build will fail is during a _release_ which is a really bad time to find problems (I don't think we push `SNAPSHOT`, `DEVEL` etc to RHEL).

Yes we don't push DEVEL or SNAPSHOT to RedHat. 
May be we should push for SNAPSHOT like we do for DockerHub so gets tested

JackPGreen:
> > We aren't just testing credentials - some of the other steps are run too, e.g. deriving HZ_VERSION, checking the yml has valid syntax etc. It's not a great test, but it's something.
> 
> Sure but main bulk of functionality is now under DRY_RUN so in reality things might still fail during release

Yes, that's a risk. This PR closes that gap (a tiny bit).


> Yes we don't push DEVEL or SNAPSHOT to RedHat. May be we should push for SNAPSHOT like we do for DockerHub so gets tested

That would help, as that would give full coverage of _actually publishing_ (missing here) - but it doesn't help validate if someone changes this action.

_Ideally_ you'd find out there's a problem with the action _at the point of changing it_, not some time later when a publish fails.

.github/workflows/tag_image_push_rhel.yml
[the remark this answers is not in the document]
nishaatr: > It doesn't actually have a default - that default value only applies when it's triggered by a workflow_dispatch. Otherwise (e.g. on push or pull_request) it's just blank.

Good to know. thanks 

Prefer we discuss further in stand-up tomorrow as not sure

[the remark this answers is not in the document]
JackPGreen: After our conversation, re-implemented.

https://github.com/hazelcast/hazelcast-docker/pull/847/commits/9002162b2a6c37f523d8458837616f3ac9bdfa96

.github/workflows/build-pr.yml
```diff
@@ -215,3 +215,24 @@ jobs:
...
+  test-push-rhel:
+    name: Test pushing RHEL image (dry run)
+    needs: [ prepare ]
+    uses: ./.github/workflows/tag_image_push_rhel.yml
+    with:
+      HZ_VERSION: ${{ needs.prepare.outputs.LAST_RELEASED_HZ_VERSION_OSS }}
+      RELEASE_VERSION: ${{ needs.prepare.outputs.LAST_RELEASED_HZ_VERSION_OSS }}
+      DRY_RUN: true
+    secrets:
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
```
nishaatr, line 228: May be use `secrets: inherit` rather than passing individual secrets?

.github/workflows/tag_image_push_rhel.yml
```diff
@@ -173,9 +225,10 @@ jobs:
 
       - name: Set MC version
         run: |
-          echo "HZ_MC_VERSION=$(echo "${{ inputs.HZ_VERSION }}" | cut -d '.' -f 1,2)" >> $GITHUB_ENV
+          echo "HZ_MC_VERSION=${{ steps.version.outputs.major }}.${{ steps.version.outputs.minor }}" >> $GITHUB_ENV
 
       - name: Deploy Hazelcast Cluster
+        if: inputs.DRY_RUN != 'true'
```
nishaatr, line 231: nit - may be set `env.DRY_RUN` at the top to force a value (so it caters for empty value in case of non-dispatch event which I presume negation fixes that) and use that `if: inputs.DRY_RUN == 'false'`
Reads better in my head and more deterministic.

nishaatr (changes requested):

.github/workflows/build-pr.yml
JackPGreen: > May be use `secrets: inherit` rather than passing individual secrets?

I don't think you can do that, if you could I would've done it that way in the other PR.

nishaatr: I saw [this ](https://docs.github.com/en/enterprise-cloud@latest/actions/sharing-automations/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow) hence the comment
May be its to do with type of workflow (i.e. reusable vs composite perhaps)

ldziedziul: > I don't think you can do that, if you could I would've done it that way in the other PR.

@JackPGreen what's the issue with `secretes: inherit` ?

.github/workflows/build-pr.yml
```diff
@@ -215,3 +215,24 @@ jobs:
...
       HZ_ENTERPRISE_LICENSE: ${{ secrets.HZ_ENTERPRISE_LICENSE }}
       SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+
+  test-push-rhel:
+    name: Test pushing RHEL image (dry run)
+    needs: [ prepare ]
+    uses: ./.github/workflows/tag_image_push_rhel.yml
+    with:
+      HZ_VERSION: ${{ needs.prepare.outputs.LAST_RELEASED_HZ_VERSION_OSS }}
+      RELEASE_VERSION: ${{ needs.prepare.outputs.LAST_RELEASED_HZ_VERSION_OSS }}
```
ldziedziul, line 225: Why explicit `RELEASE_VERSION`? It defaults to HZ_VERSION

.github/workflows/tag_image_push_rhel.yml
```diff
@@ -117,6 +166,7 @@ jobs:
           is_lts_override: ${{ inputs.IS_LTS_OVERRIDE }}
 
       - name: Build the Hazelcast Enterprise image
+        if: inputs.DRY_RUN != 'true'
```
ldziedziul, line 169: Why not build it without `--push`?

.github/workflows/build-pr.yml
JackPGreen: dd9941ef2eadfd0e1d277fd4f81b29df21774ca4

.github/workflows/tag_image_push_rhel.yml
JackPGreen: Good idea - d582a62675741dc83e26576bb5f168d3ca3b292d

.github/workflows/build-pr.yml
JackPGreen: Looks I was wrong - cool!

https://github.com/hazelcast/hazelcast-docker/pull/847/commits/1b3b4fc4710846671edac5ca25dc117f4625f4ec

.github/workflows/tag_image_push.yml
```diff
@@ -269,17 +269,22 @@ jobs:
             TAGS_ARG="${TAGS_ARG} --tag ${IMAGE_NAME}:${tag}"
           done
           
+          output=
```
ldziedziul, line 272: `output` is missleading, maybe `non_dry_run_args` ?

.github/workflows/tag_image_push_rhel.yml
ldziedziul: Great, BTW what issues have you found?

JackPGreen: > Great, BTW what issues have you found?

Added to the PR description:

> * fixes ordering issue introduced in [Add RHEL `V6` container support [DI-375] #844](https://github.com/hazelcast/hazelcast-docker/pull/844) where `RHEL_PROJECT_ID` is referenced before it's generated

.github/workflows/tag_image_push.yml
JackPGreen: I went with `output` as [that's the setting `--push` actually affects](https://docs.docker.com/reference/cli/docker/buildx/build/#push).

JackPGreen:
@nishaatr @ldziedziul are you able to review, please?

.github/workflows/tag_image_push_rhel.yml
```diff
@@ -24,6 +24,55 @@ on:
...
+      IS_LTS_OVERRIDE:
+        description: 'Override is LTS release'
+        required: false
+        type: string
+        default: ''
+      DRY_RUN:
+        description: 'Skip pushing the images to remote registry'
+        default: 'false'
+        type: string
+    secrets:
```
nishaatr, line 53: Do we need to list the secrets considering its inherited?
Looks like its needed but just checking

.github/workflows/tag_image_push_rhel.yml
```diff
@@ -135,13 +184,22 @@ jobs:
             TAGS_ARG="${TAGS_ARG} --tag ${IMAGE_NAME}:${tag}"
           done
 
+          output=
+
           PLATFORMS="linux/amd64"
-          docker buildx build --push \
+
+          if [ "${{ inputs.DRY_RUN }}" == "true" ] ; then
```
nishaatr, line 191: other places I see `if: inputs.DRY_RUN != 'true'` i.e. negated. 
thought might as well keep it consistent

.github/workflows/tag_image_push_rhel.yml
```diff
@@ -135,13 +184,22 @@ jobs:
             TAGS_ARG="${TAGS_ARG} --tag ${IMAGE_NAME}:${tag}"
           done
 
+          output=
```
nishaatr, line 187: may be rename as `args` or `options`
currently its only `--push` but we expect more options otherwise rename to `push`

nishaatr, line 187: what does `output=` actually do? In Windows this unsets the var but not sure about Bash. May be `output=""`

nishaatr (changes requested):

.github/workflows/tag_image_push_rhel.yml
nishaatr: worth doing this?

JackPGreen: > worth doing this?

Personally not a fan as we don't do it elsewhere.

.github/workflows/tag_image_push_rhel.yml
JackPGreen: Good point!

https://github.com/hazelcast/hazelcast-docker/pull/847/commits/1bea658abaaf0086be602bfe2413b6fea114c168

.github/workflows/tag_image_push_rhel.yml
JackPGreen: Already raised here - https://github.com/hazelcast/hazelcast-docker/pull/847#discussion_r1917085338

JackPGreen: It's setting a default empty string variable. `=""` might work too but elsewhere we just do `=`.

nishaatr: ok. its fine. just looked odd a bit

.github/workflows/tag_image_push_rhel.yml
ldziedziul: Check out the whole block, there's also else part. In other places we run some logic only for non-DRY_RUN, here we run two different things depending on the DRY_RUN flag

ldziedziul (approved):

.github/workflows/tag_image_push_rhel.yml
JackPGreen: > other places I see `if: inputs.DRY_RUN != 'true'` i.e. negated. thought might as well keep it consistent

Not quite sure I follow - do you mean:
- `true` / `!true`
vs
- `true` / `false`
?

nishaatr: I mean like in some steps:

```
      - name: Run preflight scan
        if: inputs.DRY_RUN != 'true'
        run: |
```
And in BASH we do 

`if [ "${{ inputs.DRY_RUN }}" == "true" ] ; then`

I am saying for consistently, lets use `inputs.DRY_RUN != 'true'` and swap the if/else logic

Not a biggy thought

JackPGreen: Sorry I still don't understand - do you have a really specific example?

nishaatr: lets skip. not a biggy :). I will approve

nishaatr (approved):

PR #859, opened 2025-01-10
Fix `openshift-tools` installation issues

JackPGreen:
In https://github.com/hazelcast/hazelcast-docker/pull/850, we use the `openshift-tools-installer` to install OpenShift tools, but [following testing](https://hazelcast.slack.com/archives/C05LM8B80UT/p1736508941644619?thread_ts=1736438770.790189&cid=C05LM8B80UT) there were some issues:

- `preflight` and `oc` must be installed from _different_ mirrors as they are [not in the same one](https://github.com/redhat-actions/openshift-tools-installer?tab=readme-ov-file#supported-tools)
- `preflight` is installed with a different name vs when simply unpacked

Testing outputs:
- [Example action showing installation success](https://github.com/hazelcast/hazelcast-docker/actions/runs/12709131411/job/35427518232)
- [`5.3.8` deployed using these changes](https://github.com/hazelcast/hazelcast-docker/actions/runs/12712052672) (via [a branch](https://github.com/hazelcast/hazelcast-docker/tree/refs/heads/5.3.8-with-%60oc%60-fix))

Post-merge actions:
- [ ] backport to maintenance branches

PR #850, opened 2025-01-09
Explicitly install OpenShift CLI tools as not included in latest GitHub Runner image [DI-396]

JackPGreen:
The latest GitHub `ubuntu-24.04` runner image [no longer includes the OpenShift CLI](https://github.com/actions/runner-images/issues/10636), which has cause [unexpected action failures](https://github.com/hazelcast/hazelcast-docker/actions/runs/12684539163):
`oc: command not found.`

Changes:
- Explicitly install the OpenShift tools [(as per `hazelcast-platform-operator`)](https://github.com/search?q=repo%3Ahazelcast%2Fhazelcast-platform-operator%20redhat-actions%2Fopenshift-tools-installer&type=code)
   - includes migrating `preflight` tool
- Replace explicit `oc login` command with action

Fixes: [DI-396](https://hazelcast.atlassian.net/browse/DI-396)

Post-merge actions:
- [ ] backport to maintenance branches

[DI-396]: https://hazelcast.atlassian.net/browse/DI-396?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #864, opened 2025-01-13
Explicitly install OpenShift CLI tools as not included in latest GitHub Runner image [DI-396] [5.5.3]

JackPGreen:
Backport of:
- https://github.com/hazelcast/hazelcast-docker/pull/850
- https://github.com/hazelcast/hazelcast-docker/pull/859


nishaatr:
Why 5.5.3?

JackPGreen:
> Why 5.5.3? Should say likely 5.5.3 will be recreated but its not confirmed yet.

It needs to be in the "live" images - else when we rebuild them, it will get deleted and not recreated. In hindsight maybe should be 5.5.2 _as well_?

nishaatr:
> > Why 5.5.3? Should say likely 5.5.3 will be recreated but its not confirmed yet.
> 
> It needs to be in the "live" images - else when we rebuild them, it will get deleted and not recreated. In hindsight maybe should be 5.5.2 _as well_?

I think lets do 5.5.2 as well. 
5.5.3 strictly not needed but no harm so I will approve.

PR #880, opened 2025-02-06
Ensure RHEL tests use a valid version of Management Center [DI-422]

JackPGreen:
The RHEL publish tests the build by connecting an Management Center instance to it.

In https://github.com/hazelcast/hazelcast-docker/pull/847 this was changed to use a _matching_ version of Hazelcast & Management Center - which isn't correct in all cases (e.g. `5.5.2` Hazelcast & `5.6.0` Management Center)

Before this change, it used the latest version of Management Center - derived by inspecting the tags of the (archived) https://github.com/hazelcast/management-center-openshift repo - but the last tag here was `5.3.2` so it wasn't really the latest at all.

In [consultation with the Management Center team](https://hazelcast.slack.com/archives/C9S7BV74Z/p1738850975557879), latest is the best option.

Changes:
- rolls back to something _inspired_ by the previous implementation - choses latest based on tags
- tags are scraped from the `hazelcast/management-center` repo
- tags are queried using the `gh` cli tool to avoid checking out the whole repo

I don't think we should backport this change, as while it's not correct in earlier versions, it doesn't cause an issues.

Fixes: [DI-422](https://hazelcast.atlassian.net/browse/DI-422)

### DIFF
--- a/.github/workflows/tag_image_push_rhel.yml
+++ b/.github/workflows/tag_image_push_rhel.yml
@@ -216,9 +216,15 @@ jobs:
 
           wait_for_container_scan "$RHEL_PROJECT_ID" "$VERSION" "$RHEL_API_KEY" "$TIMEOUT_IN_MINS"
 
-      - name: Set MC version
+      - name: Set Management Center Version to be used in the tests
         run: |
-          echo "HZ_MC_VERSION=${{ steps.version.outputs.major }}.${{ steps.version.outputs.minor }}" >> $GITHUB_ENV
+          # Find the latest, non BETA or otherwise suffix-ed tag in the repo
+          LATEST_TAG=$(gh api repos/hazelcast/management-center/tags \
+            --paginate \
+            --jq '[.[].name | select(test("^v'"${HZ_MAJOR_VERSION}"'") and (test("BETA") | not) and (test("-") | not))] | sort | last' \
+          )
+          echo ${LATEST_TAG}
+          echo "HZ_MC_VERSION=${LATEST_TAG:1}" >> $GITHUB_ENV
 
       - uses: redhat-actions/oc-login@v1
         with:
@@ -233,16 +239,16 @@ jobs:
         run: |
           WORKDIR=$(pwd)/.github/scripts
           .github/scripts/smoke-test.sh \
-                        "$WORKDIR" \
-                        "$PROJECT_NAME"  \
-                        "$SCAN_REGISTRY_USER" \
-                        "$SCAN_REGISTRY_PASSWORD" \
-                        "$SCAN_REPOSITORY" \
-                        "$RELEASE_VERSION" \
-                        "$CLUSTER_SIZE" \
-                        "$HZ_ENTERPRISE_LICENSE" \
-                        "$HZ_MC_VERSION" \
-                        "$SCAN_REGISTRY"
+                        "${WORKDIR}" \
+                        "${PROJECT_NAME}"  \
+                        "${SCAN_REGISTRY_USER}" \
+                        "${SCAN_REGISTRY_PASSWORD}" \
+                        "${SCAN_REPOSITORY}" \
+                        "${RELEASE_VERSION}" \
+                        "${CLUSTER_SIZE}" \
+                        "${HZ_ENTERPRISE_LICENSE}" \
+                        "${HZ_MC_VERSION}" \
+                        "${SCAN_REGISTRY}"
 
         env:
           CLUSTER_SIZE: 3


.github/workflows/tag_image_push_rhel.yml
```diff
@@ -216,9 +216,16 @@ jobs:
 
           wait_for_container_scan "$RHEL_PROJECT_ID" "$VERSION" "$RHEL_API_KEY" "$TIMEOUT_IN_MINS"
 
-      - name: Set MC version
+      - name: Set Management Center Version to be used in the tests
         run: |
-          echo "HZ_MC_VERSION=${{ steps.version.outputs.major }}.${{ steps.version.outputs.minor }}" >> $GITHUB_ENV
+          # Find the latest, non BETA or otherwise suffix-ed tag in the repo
+          LATEST_TAG=$(gh api repos/hazelcast/management-center/tags \
```
nishaatr, line 222: nit - `LATEST_MC_TAG`

.github/workflows/tag_image_push_rhel.yml
```diff
@@ -216,9 +216,16 @@ jobs:
...
+            --jq '[.[].name | select((test("BETA") | not) and (test("-") | not) and test("\\d+"))] | sort_by(.) | last' \
```
nishaatr, line 223: Can this filter be simpler using regex as we know the pattern?

Also, why `--jq` and not `| jq` which I think we use in other places often? Not an issue so just want to know

nishaatr (approved):

.github/workflows/tag_image_push_rhel.yml
JackPGreen: https://github.com/hazelcast/hazelcast-docker/pull/880/commits/f6e7e942c9373934d11229aa1b59fcc3a3883212

.github/workflows/tag_image_push_rhel.yml
JackPGreen: > Can this filter be simpler using regex as we know the pattern?

Maybe? Did you have something in mind?

> Also, why --jq and not | jq which I think we use in other places often? Not an issue so just want to know

You don't need to have `jq` installed to use the `--jq` functionality in the `gh` CLI - not an issue for actions but locally it's one less dependency.

Also I try to avoid pipes where possible due to catching dodgy outputs etc.

Plus generally I'd rather use the options built into tools than cludge things together - e.g. the `gh` CLI tool _could_ do that filtering server side if it wanted to.

nishaatr: > Maybe? Did you have something in mind?
> 
> >

Something like 
` echo '["v5.5.0", "test-tag","v5.4.0", "v5.5.0-BETA-1"]' | jq '.[] | select(test("^v\\d*\\.\\d*\\.\\d*$"))'`
` echo '["v5.5.0", "test-tag","v5.4.0", "v5.5.0-BETA-1"]' | jq '.[] | select(match("^v\\d*\\.\\d*\\.\\d*$"))'`

Best test it

nishaatr: > You don't need to have `jq` installed to use the `--jq` functionality in the `gh` CLI - not an issue for actions but locally it's one less dependency.
> 
> Also I try to avoid pipes where possible due to catching dodgy outputs etc.
> 
> Plus generally I'd rather use the options built into tools than cludge things together - e.g. the `gh` CLI tool _could_ do that filtering server side if it wanted to.

good points. but I think installation is not a problem as we use `jq` outside GH. 
Just wanted to know no we are good

nishaatr:
Just had a thought about sorting. Might not work with double digits e.g. `v5.5.32`
Not sure tbh but has cropped up before
MC tend to do more releases so there is chance this might happen so worth testing
you could use [sort -V](https://stackoverflow.com/questions/40390957/how-to-sort-semantic-versions-in-bash) or something similar.

JackPGreen:
> Just had a thought about sorting. Might not work with double digits e.g. `v5.5.32` Not sure tbh but has cropped up before MC tend to do more releases so there is chance this might happen so worth testing you could use [sort -V](https://stackoverflow.com/questions/40390957/how-to-sort-semantic-versions-in-bash) or something similar.

The tags are returned in date order. I've taken the sort out to avoid adding any extra logic there.

.github/workflows/tag_image_push_rhel.yml
nishaatr: great. 
`test("^v\\d+(\\.\\d+)*$"))`
I think best to test for all three parts?
Also why the grouping? is this excluding `v`?

JackPGreen: > great. `test("^v\\d+(\\.\\d+)*$"))` I think best to test for all three parts?

I don't want to _only_ find versions with three parts in case of `6.1` or `5.6.1.1` etc.

nishaatr: ah its grouping to say how many parts which makes regex shorter. nice
may be instead of `*` pin it to 2 more parts?

JackPGreen:
> > Just had a thought about sorting. Might not work with double digits e.g. `v5.5.32` Not sure tbh but has cropped up before MC tend to do more releases so there is chance this might happen so worth testing you could use [sort -V](https://stackoverflow.com/questions/40390957/how-to-sort-semantic-versions-in-bash) or something similar.
> 
> The tags are returned in date order. I've taken the sort out to avoid adding any extra logic there.

Actually changed my mind - if we do a patch release of (say) `5.1.1`, that shouldn't be considered better to use than (say) `5.6.0`. Now uses the `sort` as you suggested.

https://github.com/hazelcast/hazelcast-docker/pull/880/commits/374765ed52ee663273054e5ba03c7fdb738c44eb

nishaatr:
> Actually changed my mind - if we do a patch release of (say) `5.1.1`, that shouldn't be considered better to use than (say) `5.6.0`. Now uses the `sort` as you suggested.

Sorry where in the regex `test("^v\\d+(\\.\\d+)*$")) ` is this done i.e. should it not end with `.0`?

I guess we can limit to major/minor but not sure why we should not consider PATCH. Any good reason?
This is just for internal testing so perhaps MANOR/MINOR should be fine

JackPGreen:
> > Actually changed my mind - if we do a patch release of (say) `5.1.1`, that shouldn't be considered better to use than (say) `5.6.0`. Now uses the `sort` as you suggested.
> 
> 
> 
> Sorry where in the regex `test("^v\\d+(\\.\\d+)*$")) ` is this done i.e. should it not end with `.0`?
> 
> 
> 
> I guess we can limit to major/minor but not sure why we should not consider PATCH. Any good reason?
> 
> This is just for internal testing so perhaps MANOR/MINOR should be fine

It's not checking for 0 - that was just an example of the _kind_ of scenario where it could've gone wrong.

I was trusting GitHub returning the most recent tag would be the latest version - but when we say `latest` we really mean highest - so added an explicit semver-aware sort to assert that.

nishaatr (approved):

JackPGreen:

.github/scripts/version.functions.sh
```diff
@@ -62,3 +62,13 @@ function get_last_version_with_file() {
...
 }
+
+# Find the highest-versioned, non BETA or otherwise suffix-ed tag in the repo
+function get_last_version_tag_from_github() {
+  local repo=$1
+  tags=$(gh api repos/"${repo}"/tags \
+      --paginate \
+      --jq '.[] | select(.name | test("^v\\d+(\\.\\d+)*$")) | .name')
+  tags_sorted=$(sort --version-sort --reverse <<< "${tags}")
+  head --lines=1  <<< "${tags_sorted}"
```
ldziedziul, line 73: Why `<<<` instead of `|`?

.github/workflows/build-pr.yml
```diff
@@ -35,6 +35,8 @@ jobs:
       - name: Test scripts
         run: |
           .github/scripts/test_scripts.sh
+        env:
+          GH_TOKEN: ${{ github.token }}
```
ldziedziul, line 39: Why this change?

JackPGreen: You can't use the `gh` CLI tool unless you specify an explicit `GH_TOKEN` - and the `get_last_version_tag_from_github` implementation under test uses `gh` now.

.github/scripts/version.functions.sh
JackPGreen: I wanted to store as a temp variable for debug / `set -x` failure issues.
And _I thought_ that `<<<` here strings are better than `echo $tags_sorted | head` - I seem to remember working around some JSON escaping issue this way in the past.

PR #884, opened 2025-02-12
Fix `get-tags-to-push.sh` on OSX/ZSH

JackPGreen:
On OSX/ZSH:
```
% . .github/scripts/get-tags-to-push.sh
.github/scripts/get-tags-to-push.sh:24: parse error near `|'
```

[Escaping resolves this](https://stackoverflow.com/a/69099779).

(Covered by unit tests)

### DIFF
--- a/.github/scripts/get-tags-to-push.sh
+++ b/.github/scripts/get-tags-to-push.sh
@@ -22,7 +22,7 @@ function get_version_only_tags_to_push() {
     exit 1;
   fi
 
-  if [[ "$VERSION_TO_RELEASE" =~ .*(BETA|DEVEL).* ]]; then
+  if [[ "$VERSION_TO_RELEASE" =~ BETA|DEVEL ]]; then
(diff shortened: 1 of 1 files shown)


.github/scripts/get-tags-to-push.sh
```diff
@@ -22,7 +22,7 @@ function get_version_only_tags_to_push() {
     exit 1;
   fi
 
-  if [[ "$VERSION_TO_RELEASE" =~ .*BETA.*|.*DEVEL.* ]]; then
+  if [[ "$VERSION_TO_RELEASE" =~ .*(BETA|DEVEL).* ]]; then
```
nishaatr, line 25:
```suggestion
  if [[ "$VERSION_TO_RELEASE" =~ BETA|DEVEL ]]; then
```

nishaatr (changes requested):

nishaatr (approved):

PR #886, opened 2025-02-12
Fix `tag_image_push` for `push` triggers

JackPGreen:
The issue described in https://github.com/hazelcast/hazelcast-docker/pull/846 _also_ affects `tag_image_push`.

Specifically, when run on `push`, `RELEASE_VERSION` is empty - because it gets it value set to `${{ env.HZ_VERSION }}` which is _pre-parsed_ as an empty value and not updated when we update `HZ_VERSION` earlier.

This caused [release failures in `5.5.3`](https://github.com/hazelcast/hazelcast-docker/actions/runs/13286584060).

Post merge actions:
- [ ] backport


JackPGreen:
@ldziedziul are you able to review, please? This **will** cause issues for `5.5.4`.

PR #915, opened 2025-03-11
Fix `README` section links on DockerHub

JackPGreen:
On the [DockerHub page](https://hub.docker.com/r/hazelcast/hazelcast), links such as `Check Hazelcast Enterprise Versions for the version` do not work - because there are no corresponding anchors. Added manually.


nishaatr:
> Note that this could also have been addressed by reconfiguring the dockerhub-description action, [with caveats](https://github.com/peter-evans/dockerhub-description?tab=readme-ov-file#url-completion-known-issues).

Sorry which point is the caveat?

JackPGreen:
> > Note that this could also have been addressed by reconfiguring the dockerhub-description action, [with caveats](https://github.com/peter-evans/dockerhub-description?tab=readme-ov-file#url-completion-known-issues).
> 
> Sorry which point is the caveat?

You can use the `URL Completion` option instead, but it has "known issues".

nishaatr:
> > > Note that this could also have been addressed by reconfiguring the dockerhub-description action, [with caveats](https://github.com/peter-evans/dockerhub-description?tab=readme-ov-file#url-completion-known-issues).
> > 
> > 
> > Sorry which point is the caveat?
> 
> You can use the `URL Completion` option instead, but it has "known issues".

I presume you mean point 1 in [with caveats](https://github.com/peter-evans/dockerhub-description?tab=readme-ov-file#url-completion-known-issues)
And this seems to have the desired effect as link is converted to https://hub.docker.com/r/hazelcast/hazelcast#hazelcast-versions
Looks like the issue is that we are missing HTML anchor for section `Latest Versions` which we can add [here](https://github.com/hazelcast/hazelcast-docker/blob/master/.github/scripts/generate-docker-hub-description.sh#L39)?

JackPGreen:
> I presume you mean point 1 in [with caveats](https://github.com/peter-evans/dockerhub-description?tab=readme-ov-file#url-completion-known-issues) And this seems to have the desired effect as link is converted to https://hub.docker.com/r/hazelcast/hazelcast#hazelcast-versions

The link should go to the GitHub repo, not Dockerhub - as dockerhub doesn't have those references. Hence the link doesn't work.

> Looks like the issue is that we are missing HTML anchor for section `Latest Versions` which we can add [here](https://github.com/hazelcast/hazelcast-docker/blob/master/.github/scripts/generate-docker-hub-description.sh#L39)?

Not sure I understand?

nishaatr:
> > I presume you mean point 1 in [with caveats](https://github.com/peter-evans/dockerhub-description?tab=readme-ov-file#url-completion-known-issues) And this seems to have the desired effect as link is converted to https://hub.docker.com/r/hazelcast/hazelcast#hazelcast-versions
> 
> The link should go to the GitHub repo, not Dockerhub - as dockerhub doesn't have those references. Hence the link doesn't work.
> 
> > Looks like the issue is that we are missing HTML anchor for section `Latest Versions` which we can add [here](https://github.com/hazelcast/hazelcast-docker/blob/master/.github/scripts/generate-docker-hub-description.sh#L39)?
> 
> Not sure I understand?

Best we step back a bit:

1. the Docker link [Hazelcast Versions](https://hub.docker.com/r/hazelcast/hazelcast#hazelcast-versions) points to 'Latest Versions' in the same page (which doesn't work). End of the section says `You can find the full list of Hazelcast versions at the [Official Hazelcast Docker Hub⁠](https://store.docker.com/community/images/hazelcast/hazelcast/tags....)`
2. In GH, it says `You can find the full list of Hazelcast versions at the [Official Hazelcast Docker Hub](https://store.docker.com/community/images/hazelcast/hazelcast/tags).`

So at the end both are pointing https://store.docker.com/community/images/hazelcast/hazelcast/tags but 'Latest Versions' has some versions listed locally. If we point users to GH then they have to click again to DH. So may be should rationalize this

Option 1: make GH and DH both point to https://store.docker.com/community/images/hazelcast/hazelcast/tags and remove the list in `Latest Versions`. In fact, we don't even need section `Hazelcast Versions` and simply point to `tags` in `Check [Hazelcast Versions](https://hub.docker.com/r/hazelcast/hazelcast#hazelcast-versions)`

Option 2: just add HTML anchor in https://github.com/hazelcast/hazelcast-docker/blob/master/.github/scripts/generate-docker-hub-description.sh#L39

```
   sed -i -e "/$matching_line/ {
        a\

        a\
       <a name="hazelcast-versions"></a>
        #### Latest Versions

OR 
        #### <a name="hazelcast-versions"></a>Latest Versions
        a\
        r $tags_file
    }" "$filename"

```
I think Option 2 will work. I added the anchor in browser (Dev tools) and then the link works as is

Hope this makes sense!

nishaatr:
> Does this work via the docker action? I'd be surprised if you could add (arbitrary) HTML for security reasons.

I think so. The script first takes everything from [README.md](https://github.com/hazelcast/hazelcast-docker/blob/master/.github/scripts/generate-docker-hub-description.sh#L48) and then fill the versions
Adding HTML to markdown is allowed

> Then it wouldn't work in GitHub?

Looks like its supported https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#custom-anchors

I think worth trying!

JackPGreen:
> > Does this work via the docker action? I'd be surprised if you could add (arbitrary) HTML for security reasons.
> 
> I think so. The script first takes everything from [README.md](https://github.com/hazelcast/hazelcast-docker/blob/master/.github/scripts/generate-docker-hub-description.sh#L48) and then fill the versions Adding HTML to markdown is allowed

I was thinking more on the Dockerhub side

> > Then it wouldn't work in GitHub?
> 
> Looks like its supported https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#custom-anchors
> 
> I think worth trying!

My problem is that I can't test ^ that ^ without pushing to Dockerhub. But I can know my way works...

nishaatr:
> > > Does this work via the docker action? I'd be surprised if you could add (arbitrary) HTML for security reasons.
> > 
> > 
> > I think so. The script first takes everything from [README.md](https://github.com/hazelcast/hazelcast-docker/blob/master/.github/scripts/generate-docker-hub-description.sh#L48) and then fill the versions Adding HTML to markdown is allowed
> 
> I was thinking more on the Dockerhub side
> 
> > > Then it wouldn't work in GitHub?
> > 
> > 
> > Looks like its supported https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#custom-anchors
> > I think worth trying!
> 
> My problem is that I can't test ^ that ^ without pushing to Dockerhub. But I can know my way works...

Good point and was thinking about this. 
I see you can edit the markdown directly in DH
1. go to https://hub.docker.com/r/hazelcast/hazelcast and click `Manage Repository`
2. image
3. Go to the end of page and click `Edit`
4. image

I have added `<a name="hazelcast-versions"></a>` and it works. Check!

Use `devopshazelcast` user

JackPGreen:
> I have added `<a name="hazelcast-versions"></a>` and it works. Check!

Ok - do you want to raise a new PR for those change(s)? Or update this one instead, please?

nishaatr:
We should backport?

I think if we backport to 5.3.8 and retag it will trigger and we test that way.

nishaatr:
> > I have added `<a name="hazelcast-versions"></a>` and it works. Check!
> 
> Ok - do you want to raise a new PR for those change(s)? Or update this one instead, please?

No worries. Best here. I will get it done soon.

nishaatr:
> > We should backport?
> 
> I wasn't planning on - no-one's raised it externally, and _eventually_ it'll trickle out.

Ok. I think we should (at least 5.5.z/5) as looks a bit unprof. 
Let me do the fix first and then take it from there.

JackPGreen:
> @JackPGreen I have updated README and works. Check?

Looks good to me. Wondering if it's worth closing this PR and making a new one with an updated description now you've taken it over?

> I think the way we build DH description is not right. The DH README is not versioned so whatever the last Docker version we build is what the README will become. Perhaps this needs discussing but for now let me know what you think?

Agreed - I'd noticed the same but didn't think it was causing a problem as we rarely update the `README`. But the `IMDG` reference is a problem.

I _think_ we should rebuild the `README` when it's updated, or when a new tag is pushed - but based on `master` - https://github.com/hazelcast/hazelcast-docker/pull/918

nishaatr:
> Looks good to me. Wondering if it's worth closing this PR and making a new one with an updated description now you've taken it over?

Actually the README is from master on your branch so we can simply merge. I reverted your changes as I copied from master
[This ](https://github.com/hazelcast/hazelcast-docker/pull/915/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R319) is appearing as a change for some reason but its just a revert.
I will approve this so lets just merge it

> I think we should rebuild the README when it's updated, or when a new tag is pushed - but based on master - https://github.com/hazelcast/hazelcast-docker/pull/918

Perfect. I was going to suggest this. I will review it shortly

nishaatr (approved):

JackPGreen:
> [This ](https://github.com/hazelcast/hazelcast-docker/pull/915/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R319) is appearing as a change for some reason but its just a revert. I will approve this so lets just merge it

It's a change in line endings. Fixed in https://github.com/hazelcast/hazelcast-docker/pull/915/commits/67076ceee91b7feacfd006984f729a3c792a0043

ldziedziul (approved):

PR #918, opened 2025-03-13
Use `README` from `master` when updating Docker Hub description

JackPGreen:
As discussed in https://github.com/hazelcast/hazelcast-docker/pull/915#issuecomment-2720752695, when a tag is pushed, we update the Docker Hub description based on the `README`... in that tag. Which means when a patch for an older version is released, we effectively roll back the description to an older version.

Changes:
- on tag `push`, update the Docker Hub description in a separate job based on `master`
   - bonus - because it was in a matrix we _were_ doing it multiple times
- when the `README` or associated architecture is updated, rebuild as well
- move manual `README` updates from `tag_image_push` to the new action, which simplifies a *lot*
- convert Docker readme update to matrix to avoid duplicating action specification

Tested [here](https://github.com/hazelcast/hazelcast-docker/actions/runs/13855972414).


.github/workflows/update_readme.yml
```diff
@@ -0,0 +1,29 @@
+name: Push `README` changes to Docker Hub
+
+on:
+  push:
+    branches:
+      - master
```
nishaatr, line 6: Not sure we should do this until changes have been back ported to avoid mismatch? (which will get trigger via `.github/workflows/tag_image_push.yml`). I think this is how it should work?

1. Docker is release tagged: this will run from via `.github/workflows/tag_image_push.yml` and branch will be `master`
2. If someone wants to publish manually, they run `.github/workflows/update_readme.yml` and they can select the branch but defaults to `master`

.github/workflows/tag_image_push.yml
```diff
@@ -336,3 +329,26 @@ jobs:
...
+          ref: master
+
+      - name: Update Docker Hub Description
+        if: inputs.DRY_RUN != 'true' && (needs.prepare.outputs.should_build_readme_ee == 'yes' || needs.prepare.outputs.should_build_readme_oss == 'yes')
+        uses: ./.github/actions/update-dockerhub-description
+        with:
+          docker_username: ${{ secrets.DOCKER_USERNAME }}
+          docker_password: ${{ secrets.DOCKER_PASSWORD }}
+
+      - name: Slack notification
```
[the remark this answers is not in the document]
nishaatr: But if we call `.github/workflows/update_readme.yml` (which does Slack) then no need for Slack here

.github/workflows/update_readme.yml
```diff
@@ -0,0 +1,29 @@
+name: Push `README` changes to Docker Hub
+
+on:
```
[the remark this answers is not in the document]
JackPGreen: > Can we also add workflow_dispatch pls? Sometimes we might want to run it separately

Good point - you can run the existing `push` action manually to achieve this, but with my new changes this is unnecessarily complicated.

https://github.com/hazelcast/hazelcast-docker/pull/918/commits/fb8de84cb28f5b194abd2faf6a616d187017eb49

.github/workflows/update_readme.yml
```diff
@@ -0,0 +1,29 @@
...
+  push:
+    branches:
+      - master
+    paths:
+      - README.md
+      - .github/actions/update-dockerhub-description
+      - .github/workflows/update_readme.yml
+
+jobs:
+  push:
```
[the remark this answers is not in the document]
JackPGreen: I just copied the existing one but I guess this makes more sense

b5867e2bd8f4c23744f0648e0b78ffb9858d6b3a

.github/workflows/tag_image_push.yml
```diff
@@ -336,3 +329,26 @@ jobs:
         if: failure()
         with:
           slack-webhook-url: ${{ secrets.SLACK_WEBHOOK }}
+   
+  readme:
+    runs-on: ubuntu-latest
+    needs: prepare
```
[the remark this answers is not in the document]
JackPGreen: Good point

53455e632e5b25a463d2c53ab58d627f2b45c35a

.github/workflows/tag_image_push.yml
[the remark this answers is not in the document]
JackPGreen: Fixed as part of fcd9267c12e6f4e5cfafce2d48b030e302420d3b

.github/workflows/update_readme.yml
```diff
@@ -0,0 +1,29 @@
+name: Push `README` changes to Docker Hub
```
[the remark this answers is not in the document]
JackPGreen: We call it `Docker Hub` elsewhere so I'm keeping it consistent with that.

1cf0080bf536a67829a75b086a6c0275e6520414

.github/workflows/update_readme.yml
JackPGreen: > Not sure we should do this until changes have been back ported to avoid mismatch? (which will get trigger via `.github/workflows/tag_image_push.yml`). I think this is how it should work?
> 
> 1. Docker is release tagged: this will run from via `.github/workflows/tag_image_push.yml` and branch will be `master`
> 2. If someone wants to publish manually, they run `.github/workflows/update_readme.yml` and they can select the branch but defaults to `master`

Sorry I don't understand, especially with my recent changes. Could you have another look, please?

.github/scripts/build.functions.sh
```diff
@@ -24,25 +24,3 @@ function should_build_ee() {
     echo "no"
   fi
 }
-
-# Checks if we should rebuild the docker EE image description.
-function should_build_readme_ee() {
```
nishaatr, line 29: Glad to see the removal btw!
Enough to rely on "needs.prepare.outputs.should_build_ee == 'yes' || needs.prepare.outputs.should_build_oss"
even if README doesn't change its still ok

.github/workflows/update_readme.yml
```diff
@@ -0,0 +1,52 @@
...
+    paths:
+      - README.md
+      - .github/scripts/generate-docker-hub-description.sh
+      - .github/workflows/update_readme.yml
+
+      jobs:
+        update-readme:
+          runs-on: ubuntu-latest
+          strategy:
+            matrix:
```
nishaatr: Perfect!!

.github/workflows/update_readme.yml
nishaatr: So this will trigger after PR is merged to `master`?

If `yes` then we shouldn't do it I think because of "Not sure we should do this until changes have been back ported to avoid mismatch"

We never release `master` so doesn't make sense. We should only update description 1) release tag 2) manually and both should come from master 

That said doesn't look like the OS/EE description is version specific. But lets say we add something new then description should not be updated unless that is released. But as a rule of thumb, we should not add anything in README that is version specific as that would be confusing. Or mention version if the info is specific to a HZ version, where necessary

Does that make sense?

nishaatr (changes requested):

.github/workflows/update_readme.yml
JackPGreen: > So this will trigger after PR is merged to `master`?

Yes

> That said doesn't look like the OS/EE description is version specific. But lets say we add something new then description should not be updated unless that is released. But as a rule of thumb, we should not add anything in README that is version specific as that would be confusing. Or mention version if the info is specific to a HZ version, where necessary

Remember that there's only one `README` for all versions - e.g. we even talk about config changes required for HZ 3.x in the `README`. So if we add a new _thing_ that necessitates a `README` change, it'd _already_ need to be backwards compatible.

> then we shouldn't do it I think because of "Not sure we should do this until changes have been back ported to avoid mismatch"
> 
> We never release `master` so doesn't make sense. We should only update description 1) release tag 2) manually and both should come from master
> {...}
> Does that make sense?

Not really!

nishaatr: > Remember that there's only one README for all versions - e.g. we even talk about config changes required for HZ 3.x in the README. So if we add a new thing that necessitates a README change, it'd already need to be backwards compatible.

Lets say we change something that needs `README` change. Lets say this change is appreciable and only for 5.5.6. 
So I would make the code changes + README (mention 5.5.6) + PR + merge to `master` + backport to 5.5.z
At this point the change is not in any release as we are yet to do `5.5.6`. So the description should be updated only when we release 5.5.6. 

Makes sense? If not then let me know and I will approve. This is not a biggy. We should just be mindful when updating `README` as its now coming always from `master`

nishaatr: Yes true but I am thinking if user sees 5.5.6 then will start looking for the release and won't find it anywhere. Would look unprof IMO. We could say like `5.5.6 (to be released)` but thats a hassle. Granted this would be rare but just incase

So if we need to update description (like we have to fix the anchors), then that should be a deliberate action (either you would run after PR merge or from dev branch to test things). Otherwise, we always update description during a release; same as code changes. Publishing from `master` should always be deliberate. This makes things simpler!

JackPGreen: I understand but don’t think it’s a massive issue. If it’s a dealbreaker for you we can change it otherwise I think leave as is.

nishaatr: yes pls. Prefer we don't change anything user facing unless its deliberate
As we might not want to publish for whatever reason

JackPGreen: https://github.com/hazelcast/hazelcast-docker/pull/918/commits/f9503c1d6554dc4a9b29bb9fe00b8ac5b619789f

nishaatr: thanks

nishaatr (approved):

PR #926, opened 2025-03-14
Revert "Increase instance size used for RHEL tests [DI-435]"

JackPGreen:
Reverts hazelcast/hazelcast-docker#917

Despite testing being unable to reproduce the issue, as soon as merged [it failed in the same way as before](https://github.com/hazelcast/hazelcast-docker/actions/runs/13858020706) showing this fix does nothing.

### DIFF
--- a/.github/workflows/tag_image_push_rhel.yml
+++ b/.github/workflows/tag_image_push_rhel.yml
@@ -67,8 +67,7 @@ jobs:
       RELEASE_VERSION: ${{ inputs.RELEASE_VERSION }}
       PROJECT_NAME: test-${{ github.run_id }}-${{ github.run_attempt }}-${{ matrix.jdk }}
 
-    # Deliberately use a faster instance to avoid timeouts during smoke test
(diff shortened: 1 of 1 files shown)


nishaatr:
Why not keep for now, debug further and then see and revert if need be?
This way not need to revert backports for now. 
We don't build Docker a lot so should be ok

JackPGreen:
> Why not keep for now, debug further and then see and revert if need be?

This change offers no value to help debug, it was purely to fix:
> Waiting for test-13858020706-1-17-hazelcast-enterprise-mancenter-0 pod to be ready
error: timed out waiting for the condition on pods/test-13858020706-1-17-hazelcast-enterprise-mancenter-0

And it hasn't done that.

There's a (small) cost involved with using ubicloud.

> This way not need to revert backports for now. We don't build Docker a lot so should be ok

I wasn't planning on reverting the backports due to effort.

ldziedziul (approved):

nishaatr (approved):

PR #927, opened 2025-03-18
Remove shebang from `functions` scripts

JackPGreen:
These scripts are never called directly and instead are `source`d during other scripts executions - as such we should be trusting the parent executions shell declaration rather than overriding it.


nishaatr:
Wondering if we also need `set -euo pipefail ${RUNNER_DEBUG:+-x}`
May be enough to set in top .sh file?

nishaatr (approved):

ldziedziul (approved):

PR #930, opened 2025-04-08
Centralise supported maintenance branch definition _within `hazelcast-docker`_ [DI-349]

JackPGreen:
Extracts existing centralised version definition to a separate file so that it's not duplicated _within `hazelcast-docker`_.

We can use this a stepping-stone to de-duplicate across other repos.

Example executions showing success:
- [`check-base-images.yml`](https://github.com/hazelcast/hazelcast-docker/actions/runs/14354399113)
- [`scheduled_vulnerability_scan.yml`](https://github.com/hazelcast/hazelcast-docker/actions/runs/14333050299)

_Partially addresses_: [DI-349](https://hazelcast.atlassian.net/browse/DI-349)

[DI-349]: https://hazelcast.atlassian.net/browse/DI-349?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/check-base-images.yml
+++ b/.github/workflows/check-base-images.yml
@@ -2,6 +2,10 @@ name: Check base images
 
 on:
   workflow_dispatch:
+    inputs:
+      MINIMAL_SUPPORTED_VERSION:
+        description: 'Minimal supported version from which we should start checking images, e.g. 5.1, 5.0.1, 4.2.3. Default derived from supported maintenance versions'
+        required: false
   schedule:
     - cron: '0 6 * * *'
 
@@ -15,19 +19,22 @@ jobs:
     needs: get-maintenance-versions
     outputs:
       matrix: ${{ steps.set-matrix.outputs.matrix }}
+    env:
+      MINIMAL_SUPPORTED_VERSION: ${{ inputs.MINIMAL_SUPPORTED_VERSION }}
     steps:
       - name: Checkout Code
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
       - name: Calculate minimal supported version
-        run: echo "MINIMAL_SUPPORTED_VERSION=$(echo '${{ needs.get-maintenance-versions.outputs.versions }}' | jq '.[0]')" >> $GITHUB_ENV
+        run: echo "DEFAULT_MINIMAL_SUPPORTED_VERSION=$(echo '${{ needs.get-maintenance-versions.outputs.versions }}' | jq '.[0]')" >> $GITHUB_ENV
       - id: set-matrix
         name: Get latest patch versions
         run: |
           . .github/scripts/version.functions.sh
-          echo "Getting latest patch versions starting from ${MINIMAL_SUPPORTED_VERSION}"
-          versions=$(printf '%s\n' $(get_latest_patch_versions "${MINIMAL_SUPPORTED_VERSION}") | jq -R . | jq -c -s .)
+          MIN_VERSION=${MINIMAL_SUPPORTED_VERSION:-$DEFAULT_MINIMAL_SUPPORTED_VERSION}
+          echo "Getting latest patch versions starting from $MIN_VERSION"
+          versions=$(printf '%s\n' $(get_latest_patch_versions "${MIN_VERSION}") | jq -R . | jq -c -s .)
           echo "Found latest patch versions: $versions"
           echo "matrix={\"version\":$versions}" >> $GITHUB_OUTPUT
       - name: Slack notification
--- a/.github/workflows/check-base-images.yml
+++ b/.github/workflows/check-base-images.yml
@@ -2,22 +2,30 @@ name: Check base images
 
 on:
   workflow_dispatch:
-    inputs:
-      MINIMAL_SUPPORTED_VERSION:
-        description: 'Minimal supported version from which we should start checking images, e.g. 5.1, 5.0.1, 4.2.3. Default values is 5.3'
-        required: false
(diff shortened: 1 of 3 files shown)


.github/workflows/check-base-images.yml
```diff
@@ -2,22 +2,30 @@ name: Check base images
...
-        description: 'Minimal supported version from which we should start checking images, e.g. 5.1, 5.0.1, 4.2.3. Default values is 5.3'
-        required: false
   schedule:
     - cron: '0 6 * * *'
 
 jobs:
+  get-maintenance-versions:
+      uses: ./.github/workflows/get-supported-maintenance-versions.yaml
+
+  get-minimal-version:
```
nishaatr, line 12: May be this can be combined within `get-supported-maintenance-versions.yaml`?

nishaatr: it could also include logic for `*.*.z` branches
So its in one place and could be used somewhere else in the future without duplicating code

.github/workflows/scheduled_vulnerability_scan.yml
```diff
@@ -6,12 +6,29 @@ on:
...
+        needs: get-maintenance-versions
+        outputs:
+            branches: ${{ steps.branches.outputs.branches }}
+        steps:
+          - id: branches
+            run: |
+                # Convert maintenance versions -> Docker branches
+                maintenance_branches=$(jq --compact-output 'map(. | tostring + ".z")' <<< '${{ needs.get-maintenance-versions.outputs.versions }}')
+                # Add `master`, too
+                echo "branches=$(jq --compact-output '. as $b | ["master"] + $b' <<< "$maintenance_branches")" >> ${GITHUB_OUTPUT}
```
nishaatr, line 23: could add `master` first to array and then append `.z`?

.github/workflows/scheduled_vulnerability_scan.yml
```diff
@@ -6,12 +6,29 @@ on:
...
+
     trigger-vulnerability-scan:
+        needs: get-branches-to-scan
         name: Scan ${{ matrix.ref }}
         strategy:
             fail-fast: false
             matrix:
-                ref: [ 'master', '5.5.z', '5.4.z', '5.3.z' ]
+                ref: ${{ fromJSON(needs.get-branches-to-scan.outputs.branches) }}
```
nishaatr, line 31: Why `fromJSON`?

nishaatr (changes requested):

.github/workflows/check-base-images.yml
JackPGreen: > May be this can be combined within `get-supported-maintenance-versions.yaml`?

No - because ultimately that will get moved somewhere else.
`.z` branches are a docker-specific construct.
Plus I don't think having a minimal version actually works long term - e.g. `5.6` isn't LTS so will stop building _before_ `5.5`. But I don't want to fix everything at once.

.github/workflows/scheduled_vulnerability_scan.yml
JackPGreen: Don't have any strong opinion either way. The actual execution order of a matrix seems arbitrary in my experience. Raise a suggestion and I'll approve?

.github/workflows/scheduled_vulnerability_scan.yml
JackPGreen: > Why `fromJSON`?

Because the shared action outputs a JSON string, not an actual object. I've based it on how we centralised the JDK definition.

nishaatr: understood

.github/workflows/scheduled_vulnerability_scan.yml
nishaatr: Something like below. I thought `maintenance_branches` would be an array but its str so below does it with JQ
Try it (remove the line where `master` is added). 

`maintenance_branches=$(jq --compact-output 'map(. | tostring + ".z")  + ["master"]' <<< '${{ needs.get-maintenance-versions.outputs.versions }}')`

JackPGreen: > Something like below. I thought `maintenance_branches` would be an array but its str so below does it with JQ Try it (remove the line where `master` is added).
> 
> `maintenance_branches=$(jq --compact-output 'map(. | tostring + ".z") + ["master"]' <<< '${{ needs.get-maintenance-versions.outputs.versions }}')`

Seems to work
https://github.com/hazelcast/hazelcast-docker/pull/930/commits/1e859080ffc92f34dddd910b5e45f5d2e666d6f4

nishaatr (approved):
All good
NIT - After thought. Word `minimum` as opposed to `minimal` is more grammatical correct as per google IMO. May be `min_version` but not fussed

.github/workflows/check-base-images.yml
```diff
@@ -2,34 +2,32 @@ name: Check base images
 
 on:
   workflow_dispatch:
-    inputs:
```
ldziedziul, line 5: Without it we can't manually override the minimal version, could we keep this functionality?

JackPGreen: https://github.com/hazelcast/hazelcast-docker/pull/930/commits/dfd92d8191ebca4646022c73296db769381c4f24

PR #935, opened 2025-04-14
Support secrets in PR Builder

JackPGreen:
The [PR builder is failing](https://github.com/hazelcast/hazelcast-docker/actions/runs/14442489819/job/40495592032?pr=932) because some of the steps required credentials [not accessible to Dependabot](https://github.com/dependabot/dependabot-core/issues/3253#issuecomment-852541544).

Fixed by updating to `pull_request_target` and [using the Hazelcast guard as added to `hazelcast-code-samples`](https://github.com/hazelcast/hazelcast-code-samples/blob/master/.github/workflows/builder.yaml).

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -1,7 +1,6 @@
 name: Build PR
 
 on:
-  workflow_dispatch:
   pull_request_target:
 
 env:
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -2,7 +2,7 @@ name: Build PR
 
 on:
   workflow_dispatch:
-  pull_request:
+  pull_request_target:
 
 env:
   test_container_name_oss: hazelcast-oss-test
@@ -11,8 +11,23 @@ env:
   docker_log_file_ee: docker-hazelcast-ee-test.log
 
 jobs:
+  check_for_membership:
+    runs-on: ubuntu-latest
+    name: Check PR author membership
+    outputs:
+      check-result: ${{ steps.composite.outputs.check-result }}
+    steps: 
+      - name: Action for membership check
+        id: composite
+        uses: hazelcast/hazelcast-tpm/membership@main
+        with:
+          organization-name: 'hazelcast'
+          member-name: ${{ github.actor }}
+          token: ${{ secrets.GH_TOKEN }}
+
   prepare:
     runs-on: ubuntu-latest
+    needs: check_for_membership
     name: Prepare environment
     outputs:
       HZ_VERSION_OSS: ${{ steps.get_oss_vars.outputs.HZ_VERSION_OSS }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build-pr.yml
```diff
@@ -11,18 +11,40 @@ env:
...
     outputs:
       HZ_VERSION_OSS: ${{ steps.get_oss_vars.outputs.HZ_VERSION_OSS }}
       HZ_VERSION_EE: ${{ steps.get_ee_vars.outputs.HZ_VERSION_EE }}
       LAST_RELEASED_HZ_VERSION_OSS: ${{ steps.get_oss_vars.outputs.LAST_RELEASED_HZ_VERSION_OSS }}
       HAZELCAST_EE_ZIP_URL: ${{ steps.get_ee_vars.outputs.HAZELCAST_EE_ZIP_URL }}
     steps:
+      - name: Detect untrusted community PR
+        if: ${{ needs.check_for_membership.outputs.check-result == 'false' }}
+        run: |
+          echo "::error::ERROR: Untrusted external PR. Must be reviewed and executed by Hazelcast" 1>&2;
```
nishaatr, line 41: nit - any need for `1>&2` and why?

nishaatr (approved):

.github/workflows/build-pr.yml
JackPGreen: Good question, please ask the author of [this commit where I copied it from](https://github.com/hazelcast/hazelcast-code-samples/commit/b865852a0fb5c2271f1d56638d90bf3358d7ef2f).

ldziedziul (commented):
We can also add missing secrets here: https://github.com/hazelcast/hazelcast-docker/settings/secrets/dependabot

JackPGreen:
> We can also add missing secrets here: https://github.com/hazelcast/hazelcast-docker/settings/secrets/dependabot

That could work too, but seems more duplication.

.github/workflows/build-pr.yml
```diff
@@ -11,18 +11,40 @@ env:
...
+          exit 1
+
       - name: Checkout Code
         uses: actions/checkout@v4
         with:
+          ref: refs/pull/${{ github.event.pull_request.number }}/merge
```
ldziedziul, line 47: Does it mean that we run the workflows from base branch but with scripts and other files from a head branch?

JackPGreen: > Does it mean that we run the workflows from base branch but with scripts and other files from a head branch?

It should all be from the PR branch.

ldziedziul: Even the workflow file?

JackPGreen: The workflow file will be from `master` because it's `pull_request_target`.

ldziedziul: That's my point, the workflow will be from `master` but the rest (like scripts) from the PR branch. So if I create a new PR that changes this workflow and deletes a bash script file used here, the PR build always fail, right?

PR #936, opened 2025-04-23
Simplify RedHat smoke test

JackPGreen:
When publishing a RedHat image, we run a simple smoke test on the image:
- setup a Hazelcast cluster using our Helm chart
- pull new platform + MC images _from RedHat Registry_
- run some simple checks on the running cluster

Specifically, were overriding _some_ of the [default values in the chart's `values.yaml`](https://github.com/hazelcast/charts/blob/master/stable/hazelcast-enterprise/values.yaml) with our own - but as were doing this via multiple layers of indirection, it's not immediately obvious.

I think we can make this simpler to read by removing these layers of indirection and inlining it instead.

Note - this *does not make it more stable*.

Changes:
- `smoke-test` script inlined into calling GitHub action step
   - removes passing arguments by index
- `values.yaml` (template) + `sed` template population replaced with simply passing the arguments when we call `helm install`
   - makes it more obvious were just overriding some _default_ settings
- use default chart cluster size
   - we tell the cluster to have `3` elements and then we check it actually had `3`
   - `3` is a arbitrary and the default - we should just use the default and look it up from the chart instead
   - we shouldn't hardcode the value... twice
- use chart MC version
   - rather than looking up the latest MC version from the (private) source repo, we should simply use whatever version the chart suggests... which just so happens [to be the latest anyway](https://github.com/hazelcast/charts/blob/0de66af1b8086d71d0420b07f98f8f9dac8e9342/stable/hazelcast-enterprise/values.yaml#L349).
   - this means we can remove all the architecture to support looking the version and simplify further
      - https://github.com/hazelcast/hazelcast-docker/pull/880
      - https://github.com/hazelcast/hazelcast-docker/pull/935


.github/workflows/tag_image_push_rhel.yml
```diff
@@ -217,44 +214,45 @@ jobs:
...
-                        "${RELEASE_VERSION}" \
-                        "${CLUSTER_SIZE}" \
-                        "${HZ_ENTERPRISE_LICENSE}" \
-                        "${HZ_MC_VERSION}" \
-                        "${SCAN_REGISTRY}"
-
-        env:
-          CLUSTER_SIZE: 3
-          HZ_ENTERPRISE_LICENSE: ${{ secrets.HZ_ENTERPRISE_LICENSE }}
+          # CREATE PROJECT
```
nishaatr, line 227: We already know we want to remove the RH cluster testing
Why not just remove? (as we need this done!)

I see we are waiting on Graeme in https://hazelcast.atlassian.net/browse/DI-438
But we have been waiting for a while. May be ping him or ask QE team directly

If we are unsure or think cluster testing should be kept then we could still run it out-of-bounds.

nishaatr (changes requested):
Add JIRA in title/fixes pls

.github/scripts/cluster-verification.sh
```diff
@@ -58,11 +62,11 @@ verify_management_center() {
         else
             echo "Waiting for management center to find all ${SIZE} members..." && sleep 4
             if [ "$i" = "5" ]; then
-            echo "Management center could not find all ${SIZE} members!"
-            kubectl get pods
-            echo ""
-            kubectl logs "${PROJECT_NAME}-${NAME}-mancenter-0"
-            return 1 
+                echoerr "Management center could not find all ${SIZE} members!"
```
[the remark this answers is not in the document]
JackPGreen: Existing functionality but why not.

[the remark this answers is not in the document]
JackPGreen: https://github.com/hazelcast/hazelcast-docker/pull/936/commits/55a0993c4b63a4c7ff37ff70d206f3dbf1250288

.github/workflows/tag_image_push_rhel.yml
JackPGreen: > We already know we want to remove the RH cluster testing Why not just remove? (as we need this done!)

We aren't ready for that yet but this still offers and improvement (and scope for more in the future).

nishaatr: As discussed in refinement, this is good to go and we have additional tasks to fix testing

nishaatr (approved):

.github/workflows/build-pr.yml
```diff
@@ -1,7 +1,7 @@
 name: Build PR
 
 on:
-  pull_request_target:
+  pull_request:
```
ldziedziul, line 4: Why this change? we don't need GH_TOKEN anymore?

JackPGreen: > Why this change? we don't need GH_TOKEN anymore?

We only needed it to support the `gh` usage in the tests of the (now removed) `get_last_version_tag_from_github` function.

ldziedziul:
> [Example execution](https://github.com/hazelcast/hazelcast-docker/actions/runs/14629941297) - note the job actually failed due to (what I think is) flakiness, but shows the cluster was set up properly.

The best argument for removing those tests

ldziedziul (approved):

PR #938, opened 2025-04-24
Split redhat test

JackPGreen:


nishaatr:
Can you show that `tag_image_push_rhel.yml` passes even if test fails?

> Additionally, we run the RedHat smoke test as part of a PR

You mean test in RedHat OpenShift cluster? Can you point to the code please?

JackPGreen:
> Can you show that `tag_image_push_rhel.yml` passes even if test fails?

Not easily - I can't make it fail on demand and each execution results in a small outage (I think!) due to how it replaces the image on republish.

Bear in mind that `tag_image_push_rhel.yml` (the _workflow_) _will fail_ if the `test` _job_ fails. But it will only run after the `build` _job_ has passed, so the image won't end up going missing.

> > Additionally, we run the RedHat smoke test as part of a PR
> 
> You mean test in RedHat OpenShift cluster? Can you point to the code please?

`rhel_smoke_test` starts the cluster:
https://github.com/hazelcast/hazelcast-docker/blob/574ccaad65d2993524a35a929d9c3f8774418106/.github/workflows/rhel-smoke-test.yml#L100-L123

We are already executing `tag_image_push_rhel` as part of the PR builder:
https://github.com/hazelcast/hazelcast-docker/blob/574ccaad65d2993524a35a929d9c3f8774418106/.github/workflows/build-pr.yml#L215-L222

And you can see it running on the PR builder [here](https://github.com/hazelcast/hazelcast-docker/actions/runs/14730635537/job/41343827964).

JackPGreen:
> > Not easily - I can't make it fail on demand and each execution results in a small outage (I think!) due to how it replaces the image on republish.
> > Bear in mind that `tag_image_push_rhel.yml` (the _workflow_) _will fail_ if the `test` _job_ fails. But it will only run after the `build` _job_ has passed, so the image won't end up going missing.
> 
> IMO `tag_image_push_rhel.yml` should not fail as it will be confusing (especially during release). It should just trigger and finish regardless of the outcome. Not sure possible. May be [workflow_run ](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_run) or REST (we can use tag/branch to sync I think)

It's probably possible to trigger it in a way that it's failure won't affect anything, but if we really don't care if the test passes or fails - why bother running it at all? Until we can find a better maintainer I think this keeps it's state "monitored" but without any side effects of it failing.

> Also, we should only trigger testing if build is a SUCCESS (don't see success())

As it `needs` the `build` job I don't think it'll be triggered if that job fails.

nishaatr:
> It's probably possible to trigger it in a way that it's failure won't affect anything, but if we really don't care if the test passes or fails - why bother running it at all? Until we can find a better maintainer I think this keeps it's state "monitored" but without any side effects of it failing.

I think we do care about running the tests successfully as we decided to keep but unfortunately it has issues. So I see three options:
1. trigger as a `fire-and-forget` if its not too complicated to do
2. or, don't trigger and  run manually (we can add a line in Release instructions if need be). But this means no testing during PR
3. or forget about testing altogether. @ldziedziul voted for this but you and me wanted to keep - as at least we can say we tested it (and also we didn't have to ask if there is coverage by QE)

I think considering we have issues with testing, `Option 2` sounds reasonable for now and we can introduce testing once we manage to fix it (create JIRA).

JackPGreen:
> or, don't trigger and  run manually (we can add a line in Release instructions if need be). But this means no testing during PR

Don't forget that the image can _also_ change via scheduled rebuild (which is where most of our issues are actually occurring) - and that would be a sensible place for a smoke test.



> I think we do care about running the tests successfully as we decided to keep but unfortunately it has issues. So I see three options

I think discussions of removing the testing is part of [DI-438](https://hazelcast.atlassian.net/browse/DI-438). Here we just want it to be non-blocking for everyone else.

nishaatr:
> Don't forget that the image can _also_ change via scheduled rebuild (which is where most of our issues are actually occurring) - and that would be a sensible place for a smoke test.

Ah yes but as its flaky so will have red builds but actually shouldn't be red. Just noise we could do without. 
When I am doing release I actually check all the Actions. If I see failures then will have to dig deeper unnecessarily. I always prefer to see green builds for releases!

I would say remove for now and then take it forward with DI-438. TBH forgot about DI-438 as we both said we want to keep the testing. May be we should have concretely decided on DI-438 first and there would not have been a need to split (just remove)

JackPGreen:
> 2. or, don't trigger and  run manually (we can add a line in Release instructions if need be). But this means no testing during PR

Following our discussion, this PR has been update - the tests will no longer be (automatically) triggered.

nishaatr (approved):

ldziedziul (approved):
Could you dry run `tag_image_push_rhel` from this branch (against something than 5.5.5 (to avoid any problems with current version) ?

JackPGreen:
> Could you dry run `tag_image_push_rhel` from this branch (against something than 5.5.5 (to avoid any problems with current version) ?

https://github.com/hazelcast/hazelcast-docker/actions/runs/14752878823

PR #954, opened 2025-05-07
Add support for LTS versions [DI-258] [5.5.z]

ldziedziul:
Backport of https://github.com/hazelcast/hazelcast-docker/pull/808

Fixes: https://hazelcast.atlassian.net/browse/DI-258

The previous LTS approach had an issue that it wouldn't work when we automatically rebuild previous LTS releases (due to CVE fixes etc). It would result with two concurrent LTS builds (e.g. 5.5 and 6.5) both overwriting `latest-lts` tag. 

This PR changes the logic that before overwriting the tag we make sure that the current version is the newest LTS. It's done by looking for the latest released version containing `.github/lts` file. 

Whenever a `.github/lts` file exists the workflow will check if we're building a latest LTS version and if so it will add `latest-lts` tags and its variants (e.g. `latest-lts-jdk21`,`latest-lts-slim`, `latest-lts-slim-jdk21`, etc)

For manual runs you need to use `IS_LTS` input. It will overwrite checking the existence of the file.

Additionally added `DRY_RUN` flag to manual runs to be able test the `tag_image_push.yml` workflow in more safer way, instead of actual push you'll see something like:
```
DRY RUN: Build and push for platforms linux/arm64,linux/amd64,linux/s390x,linux/ppc64le and tags: 5.5.1 5.5.1-jdk21 5.5 5.5-jdk21 5 5-jdk21 latest latest-jdk21 latest-lts latest-lts-jdk21
```

Test runs:
- https://github.com/hazelcast/hazelcast-docker/actions/runs/14893680002
- https://github.com/hazelcast/hazelcast-docker/actions/runs/14893530645

### DIFF
--- a/.github/scripts/get-tags-to-push_tests.sh
+++ b/.github/scripts/get-tags-to-push_tests.sh
@@ -2,39 +2,23 @@
 
 set -eu
 
-function findScriptDir() {
-  CURRENT=$PWD
+SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"
 
-  DIR=$(dirname "$0")
-  cd "$DIR" || exit
-  TARGET_FILE=$(basename "$0")
-
-  # Iterate down a (possible) chain of symlinks
-  while [ -L "$TARGET_FILE" ]
-  do
-      TARGET_FILE=$(readlink "$TARGET_FILE")
-      DIR=$(dirname "$TARGET_FILE")
-      cd "$DIR" || exit
-      TARGET_FILE=$(basename "$TARGET_FILE")
-  done
-
-  SCRIPT_DIR=$(pwd -P)
-  # Restore current directory
-  cd "$CURRENT" || exit
-}
-
-findScriptDir
-
-. "$SCRIPT_DIR"/assert.sh/assert.sh
+# Source the latest version of assert.sh unit testing library and include in current shell
+assert_script_content=$(curl --silent https://raw.githubusercontent.com/hazelcast/assert.sh/main/assert.sh)
+# shellcheck source=/dev/null
+. <(echo "${assert_script_content}")
 . "$SCRIPT_DIR"/get-tags-to-push.sh
 
 TESTS_RESULT=0
 
 function assert_get_version_only_tags_to_push {
   local VERSION_TO_RELEASE=$1
-  local EXPECTED_TAGS_TO_PUSH=$2
-  local ACTUAL_TAGS_TO_PUSH=$(get_version_only_tags_to_push "$VERSION_TO_RELEASE")
-  assert_eq "$EXPECTED_TAGS_TO_PUSH" "$ACTUAL_TAGS_TO_PUSH" "Tags to push for version $VERSION_TO_RELEASE should be equal to $EXPECTED_TAGS_TO_PUSH " || TESTS_RESULT=$?
+  local IS_LTS=$2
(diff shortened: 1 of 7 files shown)


.github/scripts/get-tags-to-push_tests.sh
```diff
@@ -2,39 +2,23 @@
...
-  cd "$CURRENT" || exit
-}
-
-findScriptDir
-
-. "$SCRIPT_DIR"/assert.sh/assert.sh
+# Source the latest version of assert.sh unit testing library and include in current shell
+assert_script_content=$(curl --silent https://raw.githubusercontent.com/hazelcast/assert.sh/main/assert.sh)
+# shellcheck source=/dev/null
+. <(echo "${assert_script_content}")
```
JackPGreen, line 10: Nit - this probably didn't need to change.

.github/workflows/tag_image_push.yml
```diff
@@ -108,7 +124,7 @@ jobs:
           fetch-depth: 0
 
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@v3.1.0
+        uses: docker/setup-qemu-action@v3.2.0
```
JackPGreen, line 127: Nit - didn't need to change.

JackPGreen (approved):

nishaatr:
Add [DI-258](https://hazelcast.atlassian.net/browse/DI-258) to title

nishaatr:
> The previous LTS approach had an issue that it wouldn't work when we automatically rebuild previous LTS releases (due to CVE fixes etc). It would result with two concurrent LTS builds (e.g. 5.5 and 6.5) both overwriting latest-lts tag.

Just commenting...
Last time we discussed this we concluded that there will always only be single supported LTS, so this scenario will and should not occur. Good its fixed just in case, but conversely, I would think we should fail if we get into this scenario

ldziedziul:
> > The previous LTS approach had an issue that it wouldn't work when we automatically rebuild previous LTS releases (due to CVE fixes etc). It would result with two concurrent LTS builds (e.g. 5.5 and 6.5) both overwriting latest-lts tag.
> 
> Just commenting... Last time we discussed this we concluded that there will always only be single supported LTS, so this scenario will and should not occur. Good its fixed just in case, but conversely, I would think we should fail if we get into this scenario

Yes, there will a single version marked as LTS, but we rebuild the image of it multiple times and each we need to update lts tags

.github/workflows/tag_image_push.yml
```diff
@@ -23,6 +23,22 @@ on:
           - ALL
           - OSS
           - EE
+      IS_LTS_OVERRIDE:
```
nishaatr, line 26: Must admit don't fully understand why we have this. May be expand description or comment to explain a bit more

.github/workflows/tag_image_push.yml
```diff
@@ -23,6 +23,22 @@ on:
...
+        description: 'Override is LTS release'
+        required: false
+        type: choice
+        default: ''
```
nishaatr, line 30: why not `false`?

nishaatr (changes requested):

nishaatr:
Merged before I got chance to review 😢

JackPGreen:
> Merged before I got chance to review 😢

It's a backport of existing functionality - _if_ those changes are valid, they should be done in `master`.

PR #968, opened 2025-05-10
Fix RedHat `Image you are trying to publish did not pass the certification test` error [DI-469]

JackPGreen:
Since removing the smoke test in https://github.com/hazelcast/hazelcast-docker/pull/938, the build has become less stable due to exposing unreliable behaviour.

Specifically, when querying the state of the image in RedHat, we _assume_ that as we've deleted `unpublished` images, when we query for (say) `5.5.1-jdk17`, that _must_ be the one were trying to publish.

However, the "unpublishing" doesn't work as expected, which means the image we get when querying isn't what we expect and the rest of the process goes awry.
More specifically, an _existing_ published image (`5.5.1-jdk17`) has data _like_:
```
{
  "total": 1,
  "data": [
    {
      "repositories": [
        {
          "published": false,
          "push_date": "2025-05-10T10:39:48+00:00"
          }
        },
        {
          "published": true,
          "published_date": "2025-05-10T10:44:16.083000+00:00",
          "push_date": "2025-05-10T10:44:16.083000+00:00"
        }
      ]
    }
  ]
}
```

That's one _image_ with _both_ published and not published. So when we request to delete any unpublished images via the API, were going to delete any image that _contains_ an unpublished component - which _may_ help explain why _published_ images are going missing.

Instead, we should _explicitly_ provide the ID of the image we want to query the state for - and this sidesteps the requirement to delete unpublished images as part of the build altogether.

Changes:
- stop removing unpublished images before doing anything
- extract the `IMAGE_ID` from the preflight output
- use `IMAGE_ID` over `VERSION` (effectively the tag name)
- remove sort of `get_image` result
   - should only be one result
   - _assuming_ the first item in a list is the one you want is always going to be racey regardless of how you sort it

[Example (debug) execution](https://github.com/hazelcast/hazelcast-docker/actions/runs/14944560112/job/41986227335).

Closes: https://github.com/hazelcast/hazelcast-docker/pull/967 (would cause too many conflicts)
Fixes: [DI-469](https://hazelcast.atlassian.net/browse/DI-469)

Post-merge items:
- [ ] backport

[DI-469]: https://hazelcast.atlassian.net/browse/DI-469?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/tag_image_push_rhel.yml
+++ b/.github/workflows/tag_image_push_rhel.yml
@@ -187,21 +187,21 @@ jobs:
         run: |
           source .github/scripts/logging.functions.sh
 
-          preflight_output=$(preflight check container "${RHEL_IMAGE}" \
+          PRELIGHT_OUTPUT=$(preflight check container "${RHEL_IMAGE}" \
             --submit --pyxis-api-token=${RHEL_API_KEY} \
             --certification-component-id=${RHEL_PROJECT_ID} \
             --docker-config ~/.docker/config.json \
             2>&1)
 
-          echodebug "${preflight_output}"
+          echodebug "${PRELIGHT_OUTPUT}"
 
-          image_id=$(grep --perl-regexp --only-matching "image id is: \K[a-f0-9]+" <<< "${preflight_output}" || true)
+          image_id=$(grep --perl-regexp --only-matching "image id is: \K[a-f0-9]+" <<< "${PRELIGHT_OUTPUT}" || true)
 
           if [[ -n "${image_id}" ]]; then
             echo "IMAGE_ID=${image_id}" >> $GITHUB_ENV
           else
             echoerr "Unable to extract image ID from preflight output:"
-            echoerr "${preflight_output}"
+            echoerr "${PRELIGHT_OUTPUT}"
             exit 1
           fi
 
--- a/.github/workflows/tag_image_push_rhel.yml
+++ b/.github/workflows/tag_image_push_rhel.yml
@@ -139,14 +139,6 @@ jobs:
           release_version: ${{ env.RELEASE_VERSION }}
           is_lts_override: ${{ inputs.IS_LTS_OVERRIDE }}
 
-      - name: Delete unpublished images
-        if: inputs.DRY_RUN != 'true'
-        run: |
-          VERSION=${RELEASE_VERSION}-jdk${{ matrix.jdk }}
-          source .github/scripts/publish-rhel.sh
-
-          delete_unpublished_images "${RHEL_PROJECT_ID}" "${VERSION}" "${RHEL_API_KEY}"
-
       - name: Build the Hazelcast Enterprise image
         run: |
           . .github/scripts/get-tags-to-push.sh 
@@ -193,28 +185,40 @@ jobs:
       - name: Run preflight scan
         if: inputs.DRY_RUN != 'true'
         run: |
-          preflight check container "${RHEL_IMAGE}" \
-          --submit --pyxis-api-token=${RHEL_API_KEY} \
-          --certification-component-id=${RHEL_PROJECT_ID} \
-          --docker-config ~/.docker/config.json
+          source .github/scripts/logging.functions.sh
+
+          preflight_output=$(preflight check container "${RHEL_IMAGE}" \
+            --submit --pyxis-api-token=${RHEL_API_KEY} \
+            --certification-component-id=${RHEL_PROJECT_ID} \
+            --docker-config ~/.docker/config.json \
+            2>&1)
+
+          echo "::debug::${preflight_output}"
+
+          image_id=$(grep --perl-regexp --only-matching "image id is: \K[a-f0-9]+" <<< "${preflight_output}" || true)
+
+          if [[ -n "${image_id}" ]]; then
+            echo "IMAGE_ID=${image_id}" >> $GITHUB_ENV
+          else
+            echoerr "Unable to extract image ID from preflight output:"
+            echoerr "${preflight_output}"
+          fi
 
       - name: Wait for Scan to Complete
         if: inputs.DRY_RUN != 'true'
         run: |
-          VERSION=${RELEASE_VERSION}-jdk${{ matrix.jdk }}
           source .github/scripts/publish-rhel.sh
 
-          wait_for_container_scan "$RHEL_PROJECT_ID" "$VERSION" "$RHEL_API_KEY" "$TIMEOUT_IN_MINS"
+          wait_for_container_scan "${RHEL_PROJECT_ID}" "${IMAGE_ID}" "${RHEL_API_KEY}" "${TIMEOUT_IN_MINS}"
 
       - name: Publish the Hazelcast Enterprise image
         if: inputs.DRY_RUN != 'true'
         run: |
-          VERSION=${RELEASE_VERSION}-jdk${{ matrix.jdk }}
           source .github/scripts/publish-rhel.sh
 
-          publish_the_image "$RHEL_PROJECT_ID" "$VERSION" "$RHEL_API_KEY"
-          wait_for_container_publish "$RHEL_PROJECT_ID" "$VERSION" "$RHEL_API_KEY" "$TIMEOUT_IN_MINS"
-          sync_tags "$RHEL_PROJECT_ID" "$VERSION" "$RHEL_API_KEY"
+          publish_the_image "${RHEL_PROJECT_ID}" "${IMAGE_ID}" "${RHEL_API_KEY}"
+          wait_for_container_publish "${RHEL_PROJECT_ID}" "${IMAGE_ID}" "${RHEL_API_KEY}" "${TIMEOUT_IN_MINS}"
+          sync_tags "${RHEL_PROJECT_ID}" "${IMAGE_ID}" "${RHEL_API_KEY}"
 
       - name: Slack notification
         uses: ./.github/actions/slack-notification
--- a/.github/workflows/tag_image_push_rhel.yml
+++ b/.github/workflows/tag_image_push_rhel.yml
@@ -187,21 +187,21 @@ jobs:
         run: |
           source .github/scripts/logging.functions.sh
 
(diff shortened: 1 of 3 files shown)


nishaatr (changes requested):
Best we keep `delete_unpublished_images` logic - for housekeeping and make it easier for future debugging. Also, in RH [ticket ](https://connect.redhat.com/support/partner-acceleration-desk/#/case/04136147) they said to delete unpublished images 

The issue where image disappears during certification is a separate thing and not related  `delete_unpublished_images`. Images disappeared even before

JackPGreen:
> Best we keep `delete_unpublished_images` logic - for housekeeping and make it easier for future debugging.

I _really_ don't think this is a good idea:
- It doesn't work like how it _sounds_:
   - `Delete unpublished images`
   - Expectation
      - remove any images that haven't been published
   - Reality
      - images have _multiple_ repository items, published images can have their previous `unpublished` images, remove any image - published or not - that has one of those items
- It shouldn't be necessary to clean up a pre-state with ephemeral jobs
   - If publish of image ID `123` fails, _if_ we need to delete it were must better doing that in the teardown of the job _by ID_ than in the setup of the next execution
- Deleting stuff using a wide-ranging key is racey
- We'd need to duplicate the code to check an image status by `IMAGE_ID` and `name`

> Also, in RH [ticket ](https://connect.redhat.com/support/partner-acceleration-desk/#/case/04136147) they said to delete unpublished images

We have lots of unpublished images due to failures we've yet to fully investigate (as discussed this morning).
We should be working on stopping those failures going forwards and then cleaning up the past, rather than mopping it up every time.

> The issue where image disappears during certification is a separate thing and not related `delete_unpublished_images`. Images disappeared even before

Maybe. But they definitely will disappear when you delete them!

JackPGreen:
> > * Deleting stuff using a wide-ranging key is racey
> 
> Not strictly true. There is no issue with API per se - it always returns what ask we for (tested this numerous times). The main issue is sometimes it does not unpublish the image and that caused the script to fail.

The issue is we misinterpret the output it gives us. E.G.:

More specifically, an _existing_ published image (`5.5.1-jdk17`) has data _like_:
> 
> ```
> {
>   "total": 1,
>   "data": [
>     {
>       "repositories": [
>         {
>           "published": false,
>           "push_date": "2025-05-10T10:39:48+00:00"
>           }
>         },
>         {
>           "published": true,
>           "published_date": "2025-05-10T10:44:16.083000+00:00",
>           "push_date": "2025-05-10T10:44:16.083000+00:00"
>         }
>       ]
>     }
>   ]
> }
> ```

Based our current logic - that's an unpublished image and should be deleted. Except it's not unpublished at all - it's live! Or are you saying you think the bug on the RedHat side and the `"published": false` entry should be there?




> > If publish of image ID 123 fails, if we need to delete it were must better doing that in the teardown of the job by ID than in the setup of the next execution
> 
> I was thinking same. Can you try? (current code `delete_image` is already using image ID to delete)

Although that was a suggest of how we _could_ solve the problem, I don't think it's the _right_ solution. We should identify _why_ it keeps happening and fix it, rather than cleaning up the evidence.

nishaatr:
> Based our current logic - that's an unpublished image and should be deleted. Except it's not unpublished at all - it's live! Or are you saying you think the bug on the RedHat side and the `"published": false` entry should be there?

Yes , I think there is a bug but not the way you thinking. The JSON you posted is not complete I believe. The `repositories` block normally says `"published": false` for Quay (which we can ignore) and `true` for OpenShift when it has been successfully published
image

The issue is on RH I believe. This is my understanding

1. Lets say we have 5.3.7 as published and nothing in `unpublished` state. So we start with a clean state
2. When we request for scanning, the published image should be automatically `unpublished` (this happens because we use the same tags). But this doesn't happen every time. When we call the search command with  `published==true;` this should return nothing but not the case. 
3. At that point ` published!=true;`  should return something and it always does with `count==1` which is what we expect. Looks like every image we submit has an associated record and the record for the `published` image is not getting unpublished (or may be not deleted...not sure here)
4. The real issue is that the auto `unpublish` does not happen every time which confuses our logic. Your change to use IMAGE_ID should fix this I believe - the new IMAGE_ID will be different to the `published` one so its deterministic (I hope). 
5. Another workaround would be to remove in `wait_for_container_scan`. I tried this but then got error when actually publishing the image (timed-out)
```ruby
   if [[ ${IS_PUBLISHED} == "1" ]]; then
        echo "Image is already published, exiting"
        return 0
    fi
```

We don't have to do the cleanup but I think its better we do it. We now track the IMAGE_ID so easy to delete and should precise (need need to check first if image is truly in the `unpublished` state). As indicated, there is no real issue with the API apart from the fact that the system is not auto unpublishing the `published` image. BTW...I ran `5.3.7` by first manually unpublishing (`jdk-17` in this case) and image got published

JackPGreen:
> > Based our current logic - that's an unpublished image and should be deleted. Except it's not unpublished at all - it's live! Or are you saying you think the bug on the RedHat side and the `"published": false` entry should be there?
> 
> Yes , I think there is a bug but not the way you thinking. The JSON you posted is not complete I believe. The `repositories` block normally says `"published": false` for Quay (which we can ignore) and `true` for OpenShift when it has been successfully published image

Interesting, but I don't _think_ that changes anything.


> 2. When we request for scanning, the published image should be automatically `unpublished` (this happens because we use the same tags). But this doesn't happen every time. When we call the search command with  `published==true;` this should return nothing but not the case.

I haven't observed this behaviour, but it makes sense.



> We don't have to do the cleanup but I think its better we do it.

Struggling to understand the value in unpublishing, though. It's "wasteful" / noisy to have them there unpublished, but that's a symptom of our failure to previously investigate this problem than an issue in itself IMO. By all means we can have a manual maintenance task to remove them (although not the one I've raised with them...).

> easy to delete and should precise (need need to check first if image is truly in the `unpublished` state)

Actually we don't really, if the script fails (i.e. we have a `failure()` we post to Slack etc), we can assume _if_ we have an `IMAGE_ID`, then there's something that _could_ be cleaned up.

nishaatr:
Sure. lets leave it!
Originally wasn't too fussed on this but when I saw the comment about cleaning up in the RH ticket, I thought may be we should probably continue with the housekeeping but its not critical

.github/workflows/tag_image_push_rhel.yml
```diff
@@ -193,28 +185,40 @@ jobs:
...
-          --docker-config ~/.docker/config.json
+          source .github/scripts/logging.functions.sh
+
+          preflight_output=$(preflight check container "${RHEL_IMAGE}" \
+            --submit --pyxis-api-token=${RHEL_API_KEY} \
+            --certification-component-id=${RHEL_PROJECT_ID} \
+            --docker-config ~/.docker/config.json \
+            2>&1)
+
+          echo "::debug::${preflight_output}"
```
nishaatr, line 196: nit - can we add `echodebug` in `.github/scripts/logging.functions.sh` and use that for consistency

.github/workflows/tag_image_push_rhel.yml
```diff
@@ -193,28 +185,40 @@ jobs:
...
+
+          image_id=$(grep --perl-regexp --only-matching "image id is: \K[a-f0-9]+" <<< "${preflight_output}" || true)
+
+          if [[ -n "${image_id}" ]]; then
+            echo "IMAGE_ID=${image_id}" >> $GITHUB_ENV
+          else
+            echoerr "Unable to extract image ID from preflight output:"
+            echoerr "${preflight_output}"
```
nishaatr, line 204: will this exit?

.github/workflows/tag_image_push_rhel.yml
```diff
@@ -193,28 +185,40 @@ jobs:
       - name: Run preflight scan
         if: inputs.DRY_RUN != 'true'
         run: |
-          preflight check container "${RHEL_IMAGE}" \
-          --submit --pyxis-api-token=${RHEL_API_KEY} \
-          --certification-component-id=${RHEL_PROJECT_ID} \
-          --docker-config ~/.docker/config.json
+          source .github/scripts/logging.functions.sh
+
+          preflight_output=$(preflight check container "${RHEL_IMAGE}" \
```
nishaatr, line 190: nit - make var ALL_CAPS (like in script and easier on the eye)

.github/workflows/tag_image_push_rhel.yml
```diff
@@ -193,28 +185,40 @@ jobs:
...
+            --submit --pyxis-api-token=${RHEL_API_KEY} \
+            --certification-component-id=${RHEL_PROJECT_ID} \
+            --docker-config ~/.docker/config.json \
+            2>&1)
+
+          echo "::debug::${preflight_output}"
+
+          image_id=$(grep --perl-regexp --only-matching "image id is: \K[a-f0-9]+" <<< "${preflight_output}" || true)
```
nishaatr, line 198: nit - make var ALL_CAPS (like in script and easier on the eye)

nishaatr (changes requested):

.github/workflows/tag_image_push_rhel.yml
JackPGreen: 6bed19a5070aeb8441246110b9898a0b99e5b5c9

.github/workflows/tag_image_push_rhel.yml
JackPGreen: b3b1c1c66cebfa8b7e642a54e44b461adb05317f

.github/workflows/tag_image_push_rhel.yml
JackPGreen: fc7a0e1ce0c956406c80253cf0efc43d6ec4a2fb

nishaatr (approved):

ldziedziul (approved):
Good job!

PR #981, opened 2025-05-14
Refactor RHEL publish

JackPGreen:
The `publish_rhel` script is very long and although split out into functions, _I found_ it difficult to reason exactly what it _does_.

Refactored by:
- updated `tag_image_push_rhel` to call the script directly rather than `source`-ing them and calling _bits_ of them
- tidied up some repeated variables at the same time
- re-ordered function declarations by usage

This PR changes _nothing else_ and doesn't fix anything, it's purely a refactor.

[Example execution](https://github.com/hazelcast/hazelcast-docker/actions/runs/15019136196).

### DIFF
--- a/.github/scripts/publish-rhel.sh
+++ b/.github/scripts/publish-rhel.sh
@@ -8,9 +8,6 @@ set -o errexit -o nounset -o pipefail ${RUNNER_DEBUG:+-x}
 get_image()
 {
     local PUBLISHED=$1
-    local RHEL_PROJECT_ID=$2
-    local IMAGE_ID=$3
-    local RHEL_API_KEY=$4
 
     case "${PUBLISHED}" in
         "published")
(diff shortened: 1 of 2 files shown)


nishaatr (approved):
LGTM
Suggestion: could move `Run preflight scan` to the script as well or may be there is a reason why it was left out!

JackPGreen:
> LGTM Suggestion: could move `Run preflight scan` to the script as well or may be there is a reason why it was left out!

Don't really have a strong opinion either way, but having them as separate steps gives _some_ external visibility on where were at in the job.

nishaatr:
> > LGTM Suggestion: could move `Run preflight scan` to the script as well or may be there is a reason why it was left out!
> 
> Don't really have a strong opinion either way, but having them as separate steps gives _some_ external visibility on where were at in the job.

No strong opinion from me also but thought make sense keeping RH interaction/logic nicely contained in one place. Don't believe this will affect visibility. 
Additionally, may be we log functions with `echo "starting to..."`. I see the wait functions are missing this perhaps because it does log little later. Another option is to group the logging then it will be more clearer

But not fussed with this!

JackPGreen:

ldziedziul (approved):
LGTM, I thought, I've reviewed it before

PR #1009, opened 2025-07-07
Grant Write Permission for HZ_HOME [DEX-457]

emreyigit:
The docker image fails to write diagnostics logs on default config. The default directory for logging output is `HZ_HOME` but it's missing the write permission. Apparently, the issue was there but it's catch during [DEX-320](https://hazelcast.atlassian.net/browse/DEX-320?atlOrigin=eyJpIjoiZDkzZWJhZGZiZTQ3NGYzMDk2MGRkNjMxODdhNjc0YWMiLCJwIjoiaiJ9) verifications.

[DEX-320]: https://hazelcast.atlassian.net/browse/DEX-320?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Unfortunately, I couldn't find a way to put proper testing on Docker containers.

### DIFF
--- a/hazelcast-enterprise/Dockerfile
+++ b/hazelcast-enterprise/Dockerfile
@@ -58,8 +58,8 @@ RUN echo "Installing new packages" \
     && mv ${HZ_HOME}/tmp/*/* ${HZ_HOME}/ \
     && echo "Setting Pardot ID to 'docker'" \
     && echo 'hazelcastDownloadId=docker' > "${HZ_HOME}/lib/hazelcast-download.properties" \
-    && echo "Granting read permission to ${HZ_HOME}" \
-    && chmod -R +r ${HZ_HOME} \
+    && echo "Granting read&write&execute permissions to ${HZ_HOME}" \
+    && chmod -R a+rwX ${HZ_HOME} \
     && echo "Removing cached package data and unnecessary tools" \
     && microdnf -y remove zip unzip \
     && microdnf -y clean all \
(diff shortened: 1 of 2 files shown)


emreyigit:
Sure, @cheels is already started to test.

nishaatr:
Not sure worth adding a test or enable logging or something so this gets tested during [simple-smoke-test.sh]( https://github.com/hazelcast/hazelcast-docker/blob/master/.github/scripts/simple-smoke-test.sh)

Surprised this wasn't fixed much earlier. I mean how did it work before!

hazelcast-enterprise/Dockerfile
```diff
@@ -58,8 +58,8 @@ RUN echo "Installing new packages" \
     && mv ${HZ_HOME}/tmp/*/* ${HZ_HOME}/ \
     && echo "Setting Pardot ID to 'docker'" \
     && echo 'hazelcastDownloadId=docker' > "${HZ_HOME}/lib/hazelcast-download.properties" \
-    && echo "Granting read permission to ${HZ_HOME}" \
-    && chmod -R +r ${HZ_HOME} \
+    && echo "Granting read&write permission to ${HZ_HOME}" \
+    && chmod -R +rw ${HZ_HOME} \
```
cheels, line 62: `'w'` is not enough to allow files to be written into /opt/hazelcast directory
You have to add `chmod -R a+rwX`

emreyigit:
Another option we might override the `user.dir` property over ENV variables for container usage only and limit the permission for that folder.

cheels (approved):

ldziedziul:
> @ldziedziul diagnostics has already a default path config but it points to `user.dir` which is the `opt/hazelcast` in docker. We can't change the default path since it's a breaking change. That's why we have updated docker file.

However, we can change the path specifically within the Docker images. Currently, diagnostics doesn't work in Docker at all, so it’s hardly a breaking change.

Side note: IMHO, compromising security and stability just to avoid a "breaking change" is a very questionable decision.

PR #1013, opened 2025-07-16
Make RedHat publish serial [DI-570]

JackPGreen:
The RedHat publish is intermittently failing, most recently [here](https://github.com/hazelcast/hazelcast-docker/actions/runs/16285479358).

Specifically, the publish as asynchronous - we submit a publish request and poll for publish completion, which _intermittently_ never happens.

Further investigation [in conjunction with RedHat support](https://connect.redhat.com/support/partner-acceleration-desk/#/case/04139855) identified that the reason they aren't getting published is that _something_ is subsequently requesting the image to be _un_published. This is _probably_ a `sync-tags` request which has the (server-side) side-effect of unpublishing any subsequently-orphaned images - i.e. if an image has had it's tags removed and is no longer "accessible", it is removed.

However, although we can hypothesise _how_ they are removed, we do not understand _why_ - the images do not overlap and it _appears_ to be a race condition on the RedHat backend. This issue is more prevalent with scheduled image rebuilds where we build many versions and images in parallel.

As such, we will no longer publish RedHat images in parallel.

We have multiple layers of concurrency:
- per-JDK
   - we build JDK variants (`17`, `21` etc) of the _same_ version in parallel in the matrix
   - `max-parallel` set to restrict this
- per-version
   - multiple versions (`5.4`, `5.5`) are triggered in parallel by the scheduled image rebuild, or manually
   - `concurrency` group set to restrict this - [note that GitHub supports at most two jobs within the same concurrency group](https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs#using-concurrency-in-different-scenarios) (one running, one pending) and any additional jobs will be rejected - not queued as you might expect. This is acceptable as we _currently_ only have two versions _in_ RedHat, but _if_ that increased we may need to consider an alternative solution, additional restrictions etc.

Fixes: [DI-570](https://hazelcast.atlassian.net/browse/DI-570)

Post-merge actions:
- [ ] backport

[DI-570]: https://hazelcast.atlassian.net/browse/DI-570?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/check-base-images.yml
+++ b/.github/workflows/check-base-images.yml
@@ -49,6 +49,8 @@ jobs:
       NLC_IMAGE_NAME: ${{ secrets.NLC_IMAGE_NAME }}
     strategy:
       fail-fast: false
+      # RedHat does not support concurrent publishing, run sequentially
+      max-parallel: 1
       matrix: ${{ fromJSON(needs.get-latest-patch-versions.outputs.matrix) }}
     steps:
       - name: Checkout Code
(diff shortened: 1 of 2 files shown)


nishaatr (approved):
Nice simple fix to something I thought was a big problem and needed bigger changes 
Good one!

Wondering if we do similar thing to normal release build?

JackPGreen:
> Wondering if we do similar thing to normal release build?

This is a red hat specific issue.

nishaatr:
> > Wondering if we do similar thing to normal release build?
> 
> This is a red hat specific issue.

Understood but we also parallelize Release images. So if issue is with RH then better to serialize release builds?
Or is it the case the scheduled builds parallelism is

ldziedziul (changes requested):
> concurrency group set to restrict this - [note that GitHub supports at most two jobs within the same concurrency group](https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs#using-concurrency-in-different-scenarios) (one running, one pending) and any additional jobs will be rejected - not queued as you might expect. This is acceptable as we currently only have two versions in RedHat, but if that increased we may need to consider an alternative solution, additional restrictions etc.

I think this can be easily forgotten, and when we add another jdk the rebuild requests will be silently cancelled. 

Couldn't we wait for rhel jobs after `gh workflow run tag_image_push_rhel.yml` in `check-base-images.yml` e.g. using https://github.com/lewagon/wait-on-check-action or similar?

JackPGreen:
> 1. NLC & DockerHub can still be `workflow_dispatch` as issue is only with RH? or may be makes sense everything is `workflow_call` for consistency!

Yes but I will change it all. It's non-trivial though...

> 2. what happens if there is failure mid-way? We can rerun `.github/workflows/check-base-images.yml` I presume?

You can still re-run manually either way.

nishaatr:
> Yes but I will change it all. It's non-trivial though...

Separate PR is ok as this is fine for now!

.github/workflows/check-base-images.yml
```diff
@@ -112,21 +113,22 @@ jobs:
           else
             log_info "System packages for EE NLC ${{ matrix.version }} image are up-to-date"
           fi
-      - name: Rebuild ${{ matrix.version }} EE image
+      - name: Rebuild ${{ matrix.version }} EE RHEL image
```
ldziedziul, line 116: it's not `RHEL`

.github/workflows/tag_image_push_rhel.yml
```diff
@@ -96,8 +96,12 @@ jobs:
       RELEASE_VERSION: ${{ needs.prepare.outputs.RELEASE_VERSION }}
 
     runs-on: ubuntu-latest
+
+    # Workaround _suspected_ race conditions on the RedHat server-side publish
+    concurrency: rhel
```
ldziedziul, line 101: `concurrency: rhel` will be removed, yes?

JackPGreen: No - if we retagged a bunch of images, we still don't want to be running multiple publishes at once.

ldziedziul: but this will mean that if we add another jdk (a third one) we will silently skip the rebuild of the one jdk version

PR #1031, opened 2025-07-22
Centralise `derive-versions` logic

JackPGreen:
The logic to derive `HZ_VERSION` & `RELEASE_VERSION` is duplicated and should be consolidated.

[Example execution](https://github.com/hazelcast/hazelcast-docker/actions/runs/16457323572) (shows `RELEASE_VERSION` derived as expected, not necessarily that the tests pass).

Closes https://github.com/hazelcast/hazelcast-docker/pull/1030 (the `checkout` is no longer redundant).

### DIFF
--- a/.github/actions/derive-versions/action.yml
+++ b/.github/actions/derive-versions/action.yml
@@ -1,0 +1,36 @@
+name: Derive versions
+description: Calculates appropriate HZ_VERSION and RELEASE_VERSION
+inputs:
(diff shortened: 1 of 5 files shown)


ldziedziul (approved):
Looks good, although, after merge, I think we should think about this https://github.com/hazelcast/hazelcast-docker/pull/1012#discussion_r2205762180

PR #1036, opened 2025-07-29
Centralise Docker setup actions (qemu + buildx)

JackPGreen:
The `docker/setup-buildx-action@v3` with `version: v0.5.1` is duplicated in 7 actions. The version used for the build system should be specified only once - extracted.

This action _could_ have been externalised to `docker-actions` but seemed overengineered, instead access directly on `master` - but until merged, the builds will fail. [Example execution](https://github.com/hazelcast/hazelcast-docker/actions/runs/16607888655) when hardcoded to execute on _this_ branch.

Note - this version is [old](https://github.com/docker/buildx/releases/tag/v0.5.1) which is what prompted centralising to make it easier to explore upgrading in future.


JackPGreen:
> Good change but I think this should ideally go in `docker-actions` I believe the end goal is to have all actions there and then have minimal business logic here It will cause less confusion IMO

I thought we thought we _had_ to architect it like that to workaround GitHub issues, not that we _wanted_ to - were a long way off having only `Dockerfile`s in this repo. I'd be open to @ldziedziul opinion though.

PR #1038, opened 2025-07-29
Centralise `variants` matrix

JackPGreen:
The `variants` matrix (describing available distribution `classifiers` - either `slim` or regular) is duplicated in 4 places and can be centralised - and at the same time can centralise the computation of the `suffix` derived from the `variant`.

Also removed the `include` matrix which _appears_ to duplicate `variant` but is unused.

This action _could_ have been externalised to `docker-actions` but seemed overengineered, instead access directly on `master` - but until merged, the builds will fail.

### DIFF
--- a/.github/workflows/get-distribution-variants.yaml
+++ b/.github/workflows/get-distribution-variants.yaml
@@ -1,0 +1,17 @@
+name: Get distribution variants
+
+on:
+  workflow_call:
+    outputs:
+      variants:
+        value: ${{ jobs.get-distribution-variants.outputs.variants }}
+
+jobs:
+  get-distribution-variants:
+    runs-on: ubuntu-latest
+    outputs:
+      variants: ${{ steps.get-matrix.outputs.variants }}
+    steps:
+      - id: get-matrix
+        run: |
+          echo "variants=$(jq --null-input --compact-output '["slim", ""] | map({ classifier: ., suffix: (if . == "" then "" else "-\(.)" end) })')" >> ${GITHUB_OUTPUT}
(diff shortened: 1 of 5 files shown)


JackPGreen:
> Also this .github/workflows/get-distribution-variants.yaml should be under `/actions`

I copied [`get-supported-jdks.yaml` which is in the same place](https://github.com/hazelcast/hazelcast-docker/blob/36410467219c8215d538d4f28b4d0e2ed7a3cdb7/.github/workflows/get-supported-jdks.yaml) - I think it's because it's a "regular" workflow instead of a composite action - although I'm not sure of the pros/cons for this kind of usage.

ldziedziul:
>  instead access directly on master - but until merged, the builds will fail.

This also a reason why it makes sense to move to `docker-actions`

JackPGreen:
As discussed, moving the migration to `docker-actions` to the backlog.

.github/workflows/get-distribution-variants.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+jobs:
+  get-distribution-variants:
+    runs-on: ubuntu-latest
+    outputs:
+      variants: ${{ steps.get-matrix.outputs.variants }}
+    steps:
+      - id: get-matrix
+        run: |
+          echo "variants=$(jq --null-input --compact-output '["slim", ""] | map({ classifier: ., suffix: (if . == "" then "" else "-\(.)" end) })')" >> ${GITHUB_OUTPUT}
```
ldziedziul, line 17: Why not simply?
```suggestion
          echo "variants=[{'classifier':'slim','suffix':'-slim'},{'classifier':'','suffix':''}]" >> ${GITHUB_OUTPUT}
```

ldziedziul: Would you mind keeping it simple for now?

JackPGreen: > Would you mind keeping it simple for now?

Done.

nishaatr (approved):
LGTM

PR #1039, opened 2025-07-29
Execute `tag_image_push` actions from `master` [DI-581]

JackPGreen:
The `push` actions are automatically triggered on `push` events to their respective branches/tags, but the actions that are triggered are run on the branch themselves - which means that to support `N` maintenance versions, we end up supporting `N` copies of the build scripts.

This also exposes further issues when trying to rebuild the images on a scheduled basis - GitHub does not support dynamic usage of re-usable workflows, so instead we have to trigger the actions on the specified branches via `gh` asynchronously and then lose all visibility of the execution outcome:
https://github.com/hazelcast/hazelcast-docker/blob/36410467219c8215d538d4f28b4d0e2ed7a3cdb7/.github/workflows/check-base-images.yml#L96-L97

Update to execute from `master`, including using the latest scripts, actions etc - but to checkout the specified `ref` to use for the underlying `Dockerfile`.

Changes:
- `push` trigger removed
   - will require upstream pipeline changes
   -  actions *must* be manually triggered
- `SOURCE_REF` parameter added
   -  specify the target `hazelcast-docker` branch
- `HZ_VERSION` input removed
    - computed dynamically based on the specified branch
- optional `RELEASE_VERSION` input removed
   - allowed releasing `hazelcast-docker` with a separate lifecycle (e.g. `5.5.1.1`), outside of the release of `hazelcast` (e.g. at `5.5.1`)
   - unsure if ever been used in production, adds cognitive overhead to pipeline

[Example execution](https://github.com/hazelcast/hazelcast-docker/actions/runs/16818172462/).

Fixes: [DI-581](https://hazelcast.atlassian.net/browse/DI-581)

Pre-merge action:
- [x] re-run scheduled rebuild _after review_ to ensure it _actually works_

Post-merge action:
- [x] update maintenance branches to remove `push` trigger
- [x] update all [Jenkins Docker actions](https://jenkins.hazelcast.com/view/DevInfra/job/deploy-builds/) to use the updated branch / parameters

[DI-581]: https://hazelcast.atlassian.net/browse/DI-581?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/ee-nlc-snapshot-push.yml
```diff
@@ -3,8 +3,8 @@ name: Build EE NLC snapshot image
 on:
   workflow_dispatch:
     inputs:
-      HZ_VERSION:
-        description: 'Version of Hazelcast to build the image for'
+      REF:
```
nishaatr, line 6: REF feels too generic i.e. ref to what
While HZ_VERSION was specific
Not fussed but may be REF_BRANCH?

.github/workflows/ee-nlc-snapshot-push.yml
```diff
@@ -13,31 +13,44 @@ env:
...
-      HZ_VERSION : ${{ inputs.HZ_VERSION }}
+      HZ_VERSION: ${{ needs.prepare.outputs.HZ_VERSION }}
       NLC_IMAGE_NAME: ${{ secrets.NLC_IMAGE_NAME }}
       S3_NLC_URL: ${{ secrets.S3_NLC_URL }}
-      DOCKER_DIR: hazelcast-enterprise
+      DOCKER_DIR: ref/hazelcast-enterprise
     steps:
-      - name: Checkout Code
-        uses: actions/checkout@v5
+      - uses: actions/checkout@v5
```
nishaatr, line 48: does this need a custom folder?

.github/workflows/ee-nlc-snapshot-push.yml
```diff
@@ -13,31 +13,44 @@ env:
...
-  jdks:
+  prepare:
     runs-on: ubuntu-latest
     outputs:
+      HZ_VERSION: ${{ steps.get_hz_versions.outputs.HZ_VERSION_EE }}
       jdks: ${{ steps.jdks.outputs.jdks }}
     steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.REF }}
```
nishaatr, line 24: I see below a re-checkout under `push:`
Can we avoid this? 
Not a big problem as repo is sufficiently small and will get even smaller eventually as we move things to docker-actions but thought I check

nishaatr (changes requested):
First pass looks good with minor comments

I see `actions/checkout@v4` in some places. Not sure change here or we have dependabot PR to fix this

.github/workflows/ee-nlc-snapshot-push.yml
JackPGreen: > REF feels too generic i.e. ref to what While HZ_VERSION was specific Not fussed but may be REF_BRANCH?

I based this on the [`actions/checkout` documentation](https://github.com/actions/checkout#usage).

It could be a tag (not a branch) - or even a commit - which is why I think it's generic.

.github/workflows/ee-nlc-snapshot-push.yml
nishaatr: So `master` checks out to `<workspace>/hazelcast-docker` and `ref` to `<workspace>/ref/hazelcast-docker`?

.github/workflows/ee-nlc-snapshot-push.yml
nishaatr: I thought `prepare` could be just a step as doesn't have to be a job and do checkout first
Anyways, thought it was little wasteful but 6s is not much

JackPGreen: > I thought `prepare` could be just a step as doesn't have to be a job and do checkout first Anyways, thought it was little wasteful but 6s is not much

We run `prepare` once and then do N builds in a matrix (JDK17, JDK21, `slim` etc) using the pre-computed values.

.github/workflows/ee-nlc-snapshot-push.yml
```diff
@@ -13,31 +13,44 @@ env:
...
       matrix:
-        jdk: ${{ fromJSON(needs.jdks.outputs.jdks) }}
+        jdk: ${{ fromJSON(needs.prepare.outputs.jdks) }}
     env:
-      HZ_VERSION : ${{ inputs.HZ_VERSION }}
+      HZ_VERSION: ${{ needs.prepare.outputs.HZ_VERSION }}
       NLC_IMAGE_NAME: ${{ secrets.NLC_IMAGE_NAME }}
       S3_NLC_URL: ${{ secrets.S3_NLC_URL }}
-      DOCKER_DIR: hazelcast-enterprise
+      DOCKER_DIR: ref/hazelcast-enterprise
```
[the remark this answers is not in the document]
JackPGreen: > FTR, I see it as an error-prone, it's so easy to use wrong directory (forget to add `ref`) and haven't notice that until diverge between `master` branch and maintenance one

I understand the concern but I think the root cause is that the `push` actions are so complex (>1KLOC in this repo) that simple mistakes easy to miss.

We both have our own ways of addressing that but were getting there.

.github/workflows/ee-nlc-snapshot-push.yml
nishaatr: I think best both go to custom folder to avoid confusion and may be push changes by mistake
Having a git repo in another is not ideal

.github/workflows/tag_image_push.yml
```diff
@@ -43,17 +33,13 @@ on:
...
+        description: 'The hazelcast-docker branch to build the image from'
         required: true
-      RELEASE_VERSION:
-        type: string
-        description: 'Version to tag the docker image e.g. 5.1.1, 5.1.1-1, defaults to HZ_VERSION'
-        required: false
       RELEASE_TYPE:
         description: 'What should be built'
-        required: true
+        required: false
```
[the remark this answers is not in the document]
JackPGreen: > Why this change?

Currently, it's a required parameter, but it has a default value (for when it's missing - which it can't be) which doesn't make sense:
https://github.com/hazelcast/hazelcast-docker/blob/45ff245d0c6645369fad8896529520f359601d75/.github/workflows/tag_image_push.yml#L19-L22

In hindsight, ^ that ^ could've been raised as a separate PR before this.

When we trigger on `push`, we have no such context so assume `EE`:
https://github.com/hazelcast/hazelcast-docker/blob/45ff245d0c6645369fad8896529520f359601d75/.github/workflows/tag_image_push.yml#L79
And then derive from `release_type` file:
https://github.com/hazelcast/hazelcast-docker/blob/45ff245d0c6645369fad8896529520f359601d75/.github/workflows/tag_image_push.yml#L90-L99

Although were changing the way were triggering it, _for now_ we can use the same method for inferring the `RELEASE_TYPE`.

.github/workflows/tag_image_push.yml
```diff
@@ -74,44 +60,39 @@ jobs:
...
       should_build_ee: ${{ steps.resolve-editions.outputs.should_build_ee }}
-      HZ_VERSION: ${{ steps.derive-versions.outputs.HZ_VERSION }}
-      RELEASE_VERSION: ${{ steps.derive-versions.outputs.RELEASE_VERSION }}
+      HZ_VERSION: ${{ steps.get_hz_versions.outputs.HZ_VERSION_EE }}
       jdks: ${{ steps.jdks.outputs.jdks }}
     steps:
       - name: Checkout Code
         uses: actions/checkout@v5
+        with:
+          ref: ${{ inputs.REF }}
```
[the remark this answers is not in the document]
JackPGreen: This is correct - we don't need to checkout `master`, only `REF` so no need for a subdirectory.

We _could_ checkout to a subdirectory and then update all the following steps to account for that, but _to me_ that's more confusing.

.github/workflows/ee-nlc-snapshot-push.yml
JackPGreen: > Having a git repo in another is not ideal

We do this in lots of places already e.g.:
https://github.com/hazelcast/hazelcast-docker/blob/45ff245d0c6645369fad8896529520f359601d75/.github/workflows/check-base-images.yml#L54-L60

https://github.com/hazelcast/hazelcast-csharp-client/blob/2d6c8dbbff0d72779b495ef8677cb37c4cd0e544/.github/workflows/trigger-release.yml#L23-L44

nishaatr: personally I would keep them separate so its clear and less chance of interference/mistakes
Happy to resolve if you prefer to keep!

.github/workflows/ee-nlc-snapshot-push.yml
```diff
@@ -3,8 +3,8 @@ name: Build EE NLC snapshot image
 on:
   workflow_dispatch:
```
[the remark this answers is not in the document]
JackPGreen: I think you and @nishaatr had the [same concern](https://github.com/hazelcast/hazelcast-docker/pull/1039#discussion_r2269535093).

It's only the name of the input so if you have a better suggestion (`hazelcast-docker-source-ref`?) it's easy to change.

Of course if namings the biggest issue I feel like were getting close!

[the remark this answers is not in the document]
nishaatr: May be `RELEASE_REF`

In our case its always going to be the branch ('v' is added where needed) so may be `RELEASE_BRANCH` which will be clearer

In Jenkins https://jenkins.hazelcast.com/view/DevInfra/job/release-automation/job/distribution-automation/build?delay=0sec
image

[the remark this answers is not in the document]
JackPGreen: I think `SOURCE_REF` works. Any disagreements?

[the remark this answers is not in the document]
nishaatr: Good enough for me. Just semantics end of the day!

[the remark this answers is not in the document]
JackPGreen: https://github.com/hazelcast/hazelcast-docker/pull/1039/commits/49eb4e7c87d57d902028a572ac779a19d5d47d1d

PR #1041, opened 2025-07-30
Upgrade `buildx`

JackPGreen:
This `buildx` version is [5 years old](https://github.com/docker/buildx/releases/tag/v0.5.1) - and not managed by Dependabot.

We _could_ update to a newer specific version, but that's still not easily maintainable - instead we should just use `latest` and trust the post-build image smoke tests.

[Example execution](https://github.com/hazelcast/hazelcast-docker/actions/runs/16824160639/).

Fixes: https://github.com/hazelcast/hazelcast-docker/issues/1043

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -23,7 +23,7 @@ jobs:
       jdks: ${{ steps.jdks.outputs.jdks }}
     steps:
       - name: Checkout Code
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
         with:
           fetch-depth: 0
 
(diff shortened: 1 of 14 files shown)


nishaatr (approved):
Makes sense
I see job is stuck?
image

.github/actions/setup-docker/action.yml
```diff
@@ -8,4 +8,4 @@ runs:
     - name: Set up Docker Buildx
       uses: docker/setup-buildx-action@v3
       with:
-        version: v0.5.1
+        version: latest
```
ldziedziul, line 11: Why not remove this option and run it with default settings?

ldziedziul (approved):

PR #694, opened 2023-11-15
Generate docker images for jdk17/21 [HZ-2621]

gareth-johnston:

PR #1042, opened 2025-07-30
Support non-`amd64` platforms in RedHat

JackPGreen:
In https://github.com/hazelcast/hazelcast-docker/pull/694 (more specifically, https://github.com/hazelcast/hazelcast-docker/commit/d428943aef7e12dfd6514046c0344f96f5f1874e), the logic of assigning a `platform` to an image was centralised and managed by OS - e.g. all `ubi` images got the same set of `platform` tags.

It _appears_ as though the RHEL images were missed and therefore stuck on _only_ `amd64`.

### DIFF
--- a/.github/workflows/tag_image_push_rhel.yml
+++ b/.github/workflows/tag_image_push_rhel.yml
@@ -169,7 +169,7 @@ jobs:
 
           output=
 
-          PLATFORMS="linux/amd64"
+          PLATFORMS="$(get_ubi_supported_platforms "${{ matrix.jdk }}")"
 
           if [ "${{ inputs.DRY_RUN }}" == "true" ] ; then
             echo "DRY RUN: Skipping push for platforms ${PLATFORMS} and tags: ${TAGS_TO_PUSH}"


.github/workflows/tag_image_push_rhel.yml
```diff
@@ -169,7 +169,7 @@ jobs:
 
           output=
 
-          PLATFORMS="linux/amd64"
+          PLATFORMS="$(get_ubi_supported_platforms "${{ matrix.jdk }}")"
```
nishaatr, line 172: Looks fine but quick question 

The [get_ubi_supported_platforms](https://github.com/hazelcast/hazelcast-docker/commit/d428943aef7e12dfd6514046c0344f96f5f1874e#diff-d43e8adb3b2f0d1f020ad3d229868c02703a171638292584d762b7d0b60c3098R18) takes JDK argument but doesn't look like its being used

```shell

function get_ubi_supported_platforms() {
  local JDK=$1
  echo "linux/arm64,linux/amd64,linux/s390x,linux/ppc64le"
}
```

Are my eyes deceiving me or not needed or may be an issue?

nishaatr: didn't realize we build for `arm64`. In DockerHub I see

image

Don't see `s390x`. Wondering if list is correct?

Good spot btw!

nishaatr (changes requested):

.github/workflows/tag_image_push_rhel.yml
nishaatr: > It’s for consistency with the alpine function which does use it.

True but confusing. Why pass if JDK var is not used
Also `get_alpine_supported_platforms` seem to be declaring the platforms and it can easily call `get_ubi_supported_platforms`

nishaatr: But I guess then below would need be changed and `,linux/ppc64le` would need need to be removed instead
So no wins!
We are good in that case

```
  if [[ "$JDK" -lt 17 ]]; then
    PLATFORMS="$PLATFORMS,linux/ppc64le"
  fi
```

PR #1051, opened 2025-08-19
Execute `tag_image_push` actions from `master` [DI-581] [5.3.z]

JackPGreen:
_Partially_ backports https://github.com/hazelcast/hazelcast-docker/pull/1039 - specifically, removes the `push` triggers from the `push` actions now that they are triggered from `master`.

### DIFF
--- a/.github/workflows/tag_image_push.yml
+++ b/.github/workflows/tag_image_push.yml
@@ -1,11 +1,6 @@
 name: Build OS and EE image
 
 on:
-  push:
-    branches:
-      - "!*"
-    tags:
-      - "v5.*"
   workflow_dispatch:
     inputs:
       HZ_VERSION:


JackPGreen:
> Only 1 file changed? The other PRs all had 3

`5.3.z` has less `push` actions (e.g. RHEL) as we push to less places due to it's support status.

nishaatr:
> > > > Only 1 file changed? The other PRs all had 3
> > > 
> > > 
> > > `5.3.z` has less `push` actions (e.g. RHEL) as we push to less places due to it's support status.
> > 
> > 
> > But I think better to change just to avoid confusion
> 
> Those files don't _exist_ on this branch - we had to delete them because they would otherwise have released on `push`.

Ah yes. Of course. All good. Thanks

PR #1058, opened 2025-08-21
Avoid needlessly overwriting `HZ_VERSION`

JackPGreen:
The `Dockerfile`s have an `HZ_VERSION` arg.

We read these using [docker-actions/get-hz-versions](https://github.com/hazelcast/docker-actions/tree/master/get-hz-versions) to determine the Hazelcast version for a given `Dockerfile`.

...but then when we _build_ the `Dockerfile`, we override the `HZ_VERSION` argument with the value _we just read_ from the `Dockerfile`.

Updated to stop overriding this property.

The _only_ case when we might want to override it is in the PR builder where we build an OS image from `master`, but use a _published_ (e.g. `5.5`) version of Hazelcast... but the OS `Dockerfile` doesn't actually _use_ the `HZ_VERSION` for anything anyway so it's pointless.

### DIFF
--- a/.github/workflows/tag_image_push_rhel.yml
+++ b/.github/workflows/tag_image_push_rhel.yml
@@ -162,7 +162,9 @@ jobs:
             output=--push
           fi
 
+          # provenance disabled as causes RedHat publish failure 
           docker buildx build ${output} \
+            --provenance=false \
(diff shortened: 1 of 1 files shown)


nishaatr (approved):
Makes sense!

ldziedziul (approved):

PR #1062, opened 2025-08-23
Remove building image in `build-pr` as duplicated in `tag_image_push`

JackPGreen:
In `build-pr` we build test images and then run `simple-smoke-test`.

We subsequently trigger `tag_image_push` (in `DRY_RUN` mode) which _also_ builds a test image and then runs  `simple-smoke-test` on it (among other things).

As such we should remove this duplication, and instead rely on the script _we actually use for release_.

Post-merge actions:
- [ ] [update branch protection rules](https://github.com/hazelcast/hazelcast-docker/settings/branch_protection_rules/24406088)

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -3,23 +3,12 @@ name: Build PR
 on:
   pull_request:
 
-env:
-  test_container_name_oss: hazelcast-oss-test
-  test_container_name_ee: hazelcast-ee-test
-  docker_log_file_oss: docker-hazelcast-oss-test.log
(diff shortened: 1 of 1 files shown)


JackPGreen:
> Nice! How do want to update branch protection rules?

I’m not sure we need them - lots of our repos don’t have them and _I think_ we’re pretty good at spotting errors. And at the moment we basically dont have them anyway as most of the checks aren’t mandatory it seems.

specifically, we would need to have a mandatory check for each JDK variant passing - but hardcoding JDK17, JDK21 etc jobs isn’t scalable.

nishaatr (approved):
Nice one!

PR #1063, opened 2025-08-26
Run the tests on the production image, not test [DI-603]

JackPGreen:
In `tag_image_push.yml` we build a _test_ image, validate that and then if were happy, rebuild a _new_ image and `push` simultaneously.

Instead, we should build production image and test _locally_ and then if were happy, `push` _the same image_.

Achieved by hosting a [local registry](https://docs.docker.com/build/ci/github-actions/local-registry/) which we can `push` into and then copying it's contents externally when finished.
We must output to a registry, not the local filesystem, due to limitations of Docker + multiarch builds.

Example [execution](https://github.com/hazelcast/hazelcast-docker/actions/runs/17375602688/job/49321113779).

Fixes: [DI-603](https://hazelcast.atlassian.net/browse/DI-603)

Pre-merge actions:
- [x] _actually do_ a `push`

[DI-603]: https://hazelcast.atlassian.net/browse/DI-603?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/tag_image_push.yml
+++ b/.github/workflows/tag_image_push.yml
@@ -203,7 +203,7 @@ jobs:
             TAGS_TO_PUSH=$(get_tags_to_push "${HZ_VERSION}" "${{ matrix.variant.suffix }}" "${{ matrix.jdk }}" "${DEFAULT_JDK}" "${IS_LATEST_LTS}")
           fi
 
-          echodebug "TAGS_TO_PUSH=${TAGS_TO_PUSH}"
+          echo "TAGS_TO_PUSH=${TAGS_TO_PUSH}"
           TAGS_ARG=""
           for tag in ${TAGS_TO_PUSH[@]}
           do


.github/workflows/tag_image_push.yml
```diff
@@ -121,10 +121,15 @@ jobs:
...
     env:
-      DOCKER_ORG: hazelcast
+      LOCAL_REGISTRY: localhost:5000
+      REMOTE_REGISTRY: hazelcast
       HZ_VERSION: ${{ needs.prepare.outputs.HZ_VERSION }}
       TEST_CONTAINER_NAME: hazelcast-test
-      TEST_IMAGE_NAME: hazelcast:test
+    services:
+      registry:
+        image: registry:3
```
nishaatr, line 130: does `dependabot` update service versions?

nishaatr (approved):
Nicely done!

.github/actions/setup-docker/action.yml
```diff
@@ -7,3 +7,5 @@ runs:
 
     - name: Set up Docker Buildx
       uses: docker/setup-buildx-action@v3
+      with:
+        driver-opts: network=host
```
ldziedziul, line 11: why?

.github/workflows/tag_image_push.yml
```diff
@@ -215,7 +186,8 @@ jobs:
           TAGS_ARG=""
           for tag in ${TAGS_TO_PUSH[@]}
           do
-            TAGS_ARG="${TAGS_ARG} --tag ${IMAGE_NAME}:${tag}"
+            echo "TAG=${tag}" >> ${GITHUB_ENV}
```
ldziedziul, line 189: Why this line? You'll overwrite TAG in this loop, so only the last value will be used in the subsequent steps

.github/workflows/tag_image_push.yml
```diff
@@ -232,18 +204,43 @@ jobs:
...
+        if: inputs.DRY_RUN != 'true'
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+
+      - name: Push image
+        if: inputs.DRY_RUN != 'true'
+        run: |
+          skopeo copy --multi-arch all docker://${LOCAL_IMAGE_NAME} docker://${REMOTE_REGISTRY}/${{ matrix.distribution-type.image-name }}
```
ldziedziul, line 242: - `all` or `--all` ?
- What about all the tags we need to push?

ldziedziul (changes requested):

.github/actions/setup-docker/action.yml
JackPGreen: [It didn't work without this step](https://github.com/hazelcast/hazelcast-docker/actions/runs/17251224561/job/48953710256), and it's listed in [their example](https://docs.docker.com/build/ci/github-actions/local-registry/).

.github/workflows/tag_image_push.yml
JackPGreen: > Why this line? You'll overwrite TAG in this loop, so only the last value will be used in the subsequent steps

We need to know _one_ of the tags on the final image so we can `pull` it later - doesn't really matter which one. Previously the name was hardcoded so was a non-issue.

.github/actions/setup-docker/action.yml
JackPGreen: Add comment https://github.com/hazelcast/hazelcast-docker/pull/1063/commits/dc3a3c2e1d5f8d8ef1c1c1859f8500a3e12fd71e

.github/workflows/tag_image_push.yml
JackPGreen: I've tested the push and gotit working (albeit with a couple of changes)

https://github.com/hazelcast/hazelcast-docker/pull/1063/commits/b581b09e397d441cf81af3c718f0bff14c6b13b5

If you use `copy`, you need to copy all the tags separately, so I went for `sync` instead where you don't
But as someone else pointed out (for my own future reference!) you can workaround that by tagging the image locally with a single tag and then copying it up and doing the copy to make the new tags server-side instead.

JackPGreen: Explaining what? I think sync makes more sense anyway, as we want to copy the whole (local) repo upstream. It was more for if we want to copy between populated repos (promotion) we need to think about that.

ldziedziul (changes requested):
Looks good, but I've just noticed that I can't check what tags are meant to be pushed. 

Could you change `echodebug "TAGS_TO_PUSH=${TAGS_TO_PUSH}"` to something visible by default?

.github/workflows/tag_image_push.yml
nishaatr: understood. its fine

ldziedziul (approved):

PR #1076, opened 2025-09-01
Merge `NLC` push into `tag_image_push`

JackPGreen:
We can apply a _similar_ change from https://github.com/hazelcast/hazelcast-docker/pull/1061 to `NLC` `push`es as well.

The process for publishing `NLC` should only be different from a "regular" image in a handful of ways:
- source of the Hazelcast distribution
- requirement of `HZ_INSTANCETRACKING_FILENAME` when starting image (e.g. smoke test)
- tags

Because `NLC` relies heavily on secrets, and secrets can't be passed between jobs, refactored workflow from having configuration in the matrix to being determined inside the `push` job instead.

Post-merge actions:
- [ ] disable all the triggering jobs, as will be triggered by regular EE push
- [ ] test `push`

### DIFF
--- a/.github/workflows/tag_image_push.yml
+++ b/.github/workflows/tag_image_push.yml
@@ -194,6 +194,8 @@ jobs:
               ;;
               "nlc")
               IMAGE_NAME=${{ secrets.NLC_IMAGE_NAME }}
+              # Extract the image name from the fully-qualified destination
+              IMAGE_NAME="${IMAGE_NAME##*/}"
               ;;
               *)
               echoerr "Unknown image name for ${{ matrix.distribution-type.label }}" ; return 1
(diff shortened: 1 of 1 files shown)


JackPGreen:
Note for reviewers - the `Build and Push image` step is getting too long and should be broken down, will address separately.

JackPGreen:
> Optional: [case 1](https://github.com/hazelcast/hazelcast-docker/blob/ede8dac979108d3c17e00c9af86087eaaef1bcc8/.github/workflows/tag_image_push.yml#L232) and [case 3](https://github.com/hazelcast/hazelcast-docker/blob/ede8dac979108d3c17e00c9af86087eaaef1bcc8/.github/workflows/tag_image_push.yml#L285) could be combined or may be not!!

They probably could be, but long term should be split into separate steps (or even separate actions) to reduce complexity.

PR #1089, opened 2025-09-10
Simplify `tag_image_push_rhel` by removing `prepare` job

JackPGreen:
Since https://github.com/hazelcast/hazelcast-docker/pull/1082, this job is now triggered only after a `tag_image_push` execution in production.

This job duplicates the `prepare` job, deriving the `HZ_VERSION` etc to support _manual_ triggering via `workflow_dispatch` in the case of missing images. But we now understand images go missing during a _publish_ we are simply re-running the job in the case of the failure, the `workflow_dispatch` functionality adds no value.

Instead we can simply pass the parameters from the parent job - but we can keep `workflow_dispatch` execution (with more complicated parameters) for manual debugging if required.

### DIFF
--- a/.github/workflows/tag_image_push_rhel.yml
+++ b/.github/workflows/tag_image_push_rhel.yml
@@ -1,6 +1,17 @@
 name: Push EE RHEL image
 
 on:
+  workflow_dispatch:
+    inputs:
+      HZ_VERSION:
+        description: 'Version of Hazelcast to build the image for, e.g. 5.1.1, 5.0.1'
+        required: true
+      JDKS:
+        required: true
+      IS_LATEST_LTS:
+        required: true
+      DEFAULT_JDK:
+        required: true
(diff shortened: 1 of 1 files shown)


nishaatr (changes requested):
This means we can't directly build only RH?
Not sure if this is good or bad

JackPGreen:
> This means we can't directly build only RH?

Yes, that's it. That's why I _didn't_ do this originally... but now I don't think we need it.

nishaatr:
> > This means we can't directly build only RH?
> 
> Yes, that's it. That's why I _didn't_ do this originally... but now I don't think we need it.

can we  rely on RH to never fail in the future? not sure we can!
change does makes sense. I thought about this when reviewing the other PR but didn't mention as thought we might need it
may be provide dropdown in `tag_image_push` because have a feeling we might need it

PR #1099, opened 2025-09-21
Document `tag_image_push_rhel` parameters

JackPGreen:


.github/workflows/tag_image_push_rhel.yml
```diff
@@ -7,25 +7,33 @@ on:
         description: 'Version of Hazelcast to build the image for, e.g. 5.1.1, 5.0.1'
         required: true
       JDKS:
+        description: 'Output of `hazelcast/docker-actions/get-supported-jdks` - e.g. `[17, 21]`'
         required: true
       IS_LATEST_LTS:
+        description: 'Output of `check-if-latest-lts-release` - e.g. `false`'
         required: true
       DEFAULT_JDK:
+        description: 'Output of `docker.functions.sh`s `get_default_jdk` - e.g. `21`'
```
nishaatr, line 16: strictly speaking the 3 above vars are user inputs for `workflow_dispatch`
So little misleading I think

Also for `workflow_call`. where the values comes from are implementation detail IMO and they could be sourced from anywhere. I think best to jsut document what they are and example values

nishaatr (changes requested):

.github/workflows/tag_image_push_rhel.yml
JackPGreen: > I see. So you mean for `JDKS`, user would need to supply `[17, 21]`?

Yes:
https://github.com/hazelcast/hazelcast-docker/blob/cab865ba6fdd8f4ef024545c5a04ba03d259c323/.github/workflows/tag_image_push_rhel.yml#L10

nishaatr: must say thats little inconvenient. No issue though!
But still think its superfluous and distraction -  especially for these two (which are just self-explanatory)

```
      IS_LATEST_LTS:
        description: 'Output of `check-if-latest-lts-release` - e.g. `false`'
        required: true
      DEFAULT_JDK:
        description: 'Output of `docker.functions.sh`s `get_default_jdk` - e.g. `21`'
```

As a user don't need to know and even as `workflow_call` also. The important bit is what the input is and the format. 
This way if those functions change then no need to update the doc.

JackPGreen: > > You only want the non-trivial params documented?
> 
> Actually yes, lets do that. So leave `JDKS` as is and simplify `IS_LATEST_LTS` and `DEFAULT_JDK` as they are no brainers

https://github.com/hazelcast/hazelcast-docker/pull/1099/commits/f5cea908d34d38fdcb631073033cf7acb2f3cb37

nishaatr (approved):

.github/workflows/tag_image_push_rhel.yml
```diff
@@ -7,19 +7,23 @@ on:
...
         required: true
       JDKS:
+        description: 'Output of `hazelcast/docker-actions/get-supported-jdks` - e.g. `[17, 21]`'
         required: true
       IS_LATEST_LTS:
         required: true
       DEFAULT_JDK:
         required: true
       ENVIRONMENT:
+        description: 'Environment to use'
```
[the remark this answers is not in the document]
JackPGreen: > Maybe add choices here to have a dropdown menu?

Eventually (soon!) this will be dynamic.
Plus needing to add code to support a new externalised config option isn’t great

.github/workflows/tag_image_push_rhel.yml
ldziedziul: > Actually yes, lets do that. So leave `JDKS` as is and simplify `IS_LATEST_LTS` and `DEFAULT_JDK` as they are no brainers

I don't agree, they should have description with example values, otherwise one could think:
- `IS_LATEST_LTS` - could be `true/false`, `yes/no`, `0/1`
- `DEFAULT_JDK` - could be `21`, `21.0`, `21-amzn`, `25.0.0.36.2`

PR #1094, opened 2025-09-16
Simplify NLC secrets

JackPGreen:
Simplify NLC secrets
NLC has two overlapping secrets (example Docker _equivalent_ values for clarity):
- `NLC_REPOSITORY` - registry (`docker.io`)
- `NLC_IMAGE_NAME` - registry/organisation/image-name (`docker.io/hazelcast/hazelcast-enterprise`)

Issues:
- duplicating the registry offers no value
- in `ee-nlc-tag-push` we do string manipulation to extract `image-name` from `NLC_IMAGE_NAME`
- the organisation and image-name have been hardcoded (i.e. leaked) in `test_published_images.yml`

Changes:
- update `NLC_IMAGE_NAME` to refer to the... image name
- un-secret the organisation name

Simplifying these secrets will make it easier when we come to migrate them to being environment-specific.

Post-merge actions:
- [x] update `NLC_IMAGE_NAME` secret

PR #1109, opened 2025-09-23
Externalise registry namespaces

JackPGreen:
Rather than using hardcoded namespaces (i.e. `hazelcast` org in Dockerhub), these should be externalised to the environment.

`hazelcast` can refer to many things and it's easy to get confused (e.g. image name, directory, namespace, artifact, GitHub org etc). As these aren't a secret, I've made them _variables_ so they print nicely in logs.

Follow on from [PR feedback](https://github.com/hazelcast/hazelcast-docker/pull/1094#issuecomment-3317875199).

### DIFF
--- a/.github/workflows/ee-nlc-tag-push.yml
+++ b/.github/workflows/ee-nlc-tag-push.yml
@@ -200,10 +200,6 @@ jobs:
       - name: Push image to remote registry
         if: inputs.DRY_RUN != 'true'
         run: |
-          remote_registry=${{ secrets.NLC_REPOSITORY }}/${{ vars.NLC_NAMESPACE }}/${{ secrets.NLC_IMAGE_NAME }}
-          # Extract the registry from the fully-qualified destination
-          remote_registry="${remote_registry%/*}"
-
           # Skopeo expects HTTPS registry but local is HTTP
           skopeo sync \
             --all \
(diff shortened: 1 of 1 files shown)


nishaatr (approved):
Personally I would have preferred ORG over NAMESPACE
Unlike for Maven Central, it has NAMESPACE as its a shared resource whilst DockerHub, RH etc have dedicated repos in an org. NAMESPACE also conjures up 
namespace in languages. Feels like we are coercing unnecessarily
Anyways, not too fussed!

nishaatr:
> > Personally I would have preferred ORG over NAMESPACE
> 
> `namespace` is their terminology (although they aren't 100% consistent...)
> 
> Example - https://docs.docker.com/reference/api/hub/latest/#tag/repositories/operation/CreateRepository

Didn't know.

nishaatr:
> > > Personally I would have preferred ORG over NAMESPACE
> > 
> > 
> > `namespace` is their terminology (although they aren't 100% consistent...)
> > Example - https://docs.docker.com/reference/api/hub/latest/#tag/repositories/operation/CreateRepository
> 
> Didn't know.

FYI...
Found this https://stevelasker.blog/2020/02/17/registry-namespace-repo-names/
had cursory look but seems to encourage use of `namespece`

PR #1214, opened 2026-01-23
Remove `workflow_dispatch` / `workflow_call` `inputs` duplication with YAML anchors

JackPGreen:
Using ([newly supported](https://github.blog/changelog/2025-09-18-actions-yaml-anchors-and-non-public-workflow-templates/)) [YAML anchors](https://docs.github.com/en/actions/reference/workflows-and-actions/reusing-workflow-configurations#yaml-anchors-and-aliases) we can avoid duplicating inputs between `workflow_dispatch` and `workflow_call` triggers - when the inputs are identical.

This isn't _always_ possible as they support different types (e.g. [`workflow_dispatch` supports `choice`](https://docs.github.com/en/enterprise-cloud@latest/actions/reference/workflows-and-actions/workflow-syntax#onworkflow_dispatchinputsinput_idtype), [`workflow_call` does not](https://docs.github.com/en/enterprise-cloud@latest/actions/reference/workflows-and-actions/workflow-syntax#onworkflow_callinputsinput_idtype)).

### DIFF
--- a/.github/workflows/ee-nlc-tag-push.yml
+++ b/.github/workflows/ee-nlc-tag-push.yml
@@ -3,26 +3,22 @@ name: Build EE NLC image
 on:
   workflow_dispatch:
     inputs:
-      SOURCE_REF:
+      SOURCE_REF: &INPUT_SOURCE_REF
         description: 'The hazelcast-docker branch to build the image from'
         required: true
-      HZ_REVISION:
+        type: string
+      HZ_REVISION: &INPUT_HZ_REVISION
         description: 'Commit id of Hazelcast snapshot jar'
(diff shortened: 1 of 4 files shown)


ldziedziul (commented):
> > That said Google did suggest `yq` which comes bundled in GH runner **(just FYI and not suggesting we do this)**
> 
> Sorry, I'm confused on the application of `yq` here. Are you suggesting we add some linter that detects changes to workflows and then keeps the triggers in sync between them?

I have mixed opinion on using these anchors, it makes reading harding, as I have two jump between two triggers field by field. Maybe a linter or test would be a better choice

PR #1268, opened 2026-04-20
Remove GitHub Release Creation

JackPGreen:
We use a third-party GitHub Action to create GitHub releases - but what value do they add?

The repo rebuilds images regularly using dynamic content, and we even retag when backporting fixes - there's no fixed "release" for us to anchor against to add value.


ldziedziul (approved):
I think watchers get notifications when there's a new release, but no fixed-tag releases make it only confusing

PR #1279, opened 2026-05-29
Use GitHub `environment` without `deployment` for non-deploying workflows

JackPGreen:
See:
- https://github.blog/changelog/2026-03-19-github-actions-late-march-2026-updates/#github-actions-now-allows-developers-to-use-environments-without-auto-deployment
- https://docs.github.com/en/actions/how-tos/deploy/configure-and-manage-deployments/control-deployments#using-environments-without-deployments

We want to avoid endless `temporary deployed` notices for workflows (eg https://github.com/hazelcast/hazelcast-docker/pull/1278) that don't deploy anything as it's just confusing.


ldziedziul (approved):
OMG! ❤️

PR #1286, opened 2026-07-10
Migrate Docker release version updates out of pipeline

JackPGreen:
The `Dockerfile`s start off with `SNAPSHOT` versions:
https://github.com/hazelcast/hazelcast-docker/blame/2016f509c9efd8976f0ef546f1d8cc589545880e/hazelcast-enterprise/Dockerfile#L3

But before release, the pipeline migrates those to release versions before the release itself is triggered.
We can simplify by moving this update into the action, too.

[Sandbox execution](https://github.com/hz-devops-test/hazelcast-docker/actions/runs/29088665218), compare the [previous pipeline commit](https://github.com/hz-devops-test/hazelcast-docker/commit/46b27d521163daa130242ef97e2a193a17828023) to the [new one created by the action](https://github.com/hz-devops-test/hazelcast-docker/commit/85d710d7f0b56403228ea4976563967738e4189a).

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -56,7 +56,7 @@ jobs:
         with:
           ref: ${{ inputs.VERSION }}
 
-      - name: Update `SNAPSHOT` to release version
+      - name: Update Dockerfile `SNAPSHOT` version to release version
         run: |
           find . -type f -iname 'Dockerfile' \
             -exec sed -i -E "s/^ARG HZ_VERSION=.*/ARG HZ_VERSION=${VERSION}/" {} +
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -39,9 +39,12 @@ on:
         type: string
 
 jobs:
-  log-distinct-id:
+  update-version:
     runs-on: ubuntu-slim
-    if: inputs.distinct_id != ''
+    permissions:
+      contents: write
+    env:
+      VERSION: ${{ inputs.VERSION }}
     steps:
         # https://github.com/Codex-/return-dispatch#receiving-repository-action
       - name: Logging distinct ID (${{ inputs.distinct_id }})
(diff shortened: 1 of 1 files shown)


.github/workflows/package.yml
```diff
@@ -39,17 +39,38 @@ on:
...
         env:
           DISTINCT_ID: ${{ inputs.distinct_id }}
 
+      - uses: actions/checkout@v7
+        with:
+          ref: ${{ inputs.VERSION }}
+
+      - name: Update `SNAPSHOT` to release version
+        run: |
+          find . -type f -iname 'Dockerfile' \
```
nishaatr, line 52: add comment to say updates OSS & EE Dockerfiles

.github/workflows/package.yml
```diff
@@ -39,17 +39,38 @@ on:
...
+            -exec sed -i -E "s/^ARG HZ_VERSION=.*/ARG HZ_VERSION=${VERSION}/" {} +
+
+      - name: Commit
```
nishaatr, line 55: No PR i.e. `package/promote`?

nishaatr (changes requested):

.github/workflows/package.yml
JackPGreen: > add comment to say updates OSS & EE Dockerfiles

c4ce05644e852db13fd7c900cbbf2b4a17a876c5

.github/workflows/package.yml
JackPGreen: Good point, we could do it with PRs but gets very complicated.

The justification for PRs in other repos is twofold:
- that we want an atomic release - i.e. when we merge `rel-scripts` or `homebrew-hz`, that is actually releasing an artifact into the wild
- to perform additional validation that the change doesn't break `master` for everyone else

Here we are just updating the version number in a pre-existing (public) branch. We aren't releasing anything with this change - although _ideally_ we'd avoid any changes until we are ready to promote.

However, we do need this change (the Dockerfile to be updated) before we can build the images. So we could poll-and-wait for the PR to be merged, or instead raise a PR and have all the downstream actions use the PR branch rather than the triggering branch and then merge the PR when we promote.

I don't think either is necessary here.

nishaatr: understood.
last point came to mind so agreed

nishaatr (approved):

ldziedziul (approved):

nishaatr (approved):

PR #1292, opened 2026-07-30
Always authenticate with Dockerhub when _building_ images

JackPGreen:
[A build failed](https://github.com/hazelcast/hazelcast-docker/actions/runs/30504812595/job/90752173509#step:11:207):
> #4 ERROR: failed to authorize: failed to fetch oauth token: Post "https://auth.docker.io/token": read tcp 10.1.0.114:53532->172.64.144.78:443: read: connection reset by peer

While the error is cryptic, [Dockerhub has rate limits](https://docs.docker.com/docker-hub/usage/) which _might_ be causing the issue.

Specifically - whenever we build an image, we implicitly pull the base image from Dockerhub - in the case of NLC builds, as we never _push_ to Dockerhub, we don't authenticate and therefore these pulls _might_ be unauthenticated.

Updated to _always_ authenticate with Dockerhub when building.

### DIFF
--- a/.github/actions/setup-docker/action.yml
+++ b/.github/actions/setup-docker/action.yml
@@ -1,4 +1,5 @@
 name: Setup Docker
+description: Configures runner with tools required to build (not just manage) Docker images 
 inputs:
   platforms:
     description: Platforms to install (e.g., `arm64,riscv64,arm`)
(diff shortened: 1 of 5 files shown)


nishaatr (changes requested):
Not clear why NLC reaches out to DockerHub?

JackPGreen:
> Not clear why NLC reaches out to DockerHub?

It needs to pull the base image _from_ docker hub to build on top of.

nishaatr (approved):
May be use OIDC at the back of https://github.com/hazelcast/hazelcast-docker/pull/1293
Sonar needs addressing!

JackPGreen:
For my own reference - it seems maybe this was not required after all - https://github.com/docker/login-action/issues/862#issuecomment-2991979855

PR #1293, opened 2026-07-30
Authenticate with Dockerhub via OIDC [DI-847]

JackPGreen:
See:
- https://github.com/docker/login-action#docker-hub
- https://docs.docker.com/enterprise/security/oidc-connections/

[Support is limited to Docker registries](https://github.com/docker/oidc-action#docker-oidc-action), so API interactivity still uses password.


nishaatr (approved):
LGTM

Looks like you created OIDC connection in DockerHub
worth adding that in description as had to check

PR #1296, opened 2026-08-05
Move to Hazelcast 6.0 development cycle

nishaatr:
At the back of https://github.com/hazelcast/hazelcast-mono/pull/7393
Normally, the Release Pipeline does this update during release branch creation
But we have agreed to update manually

### DIFF
--- a/hazelcast-enterprise/Dockerfile
+++ b/hazelcast-enterprise/Dockerfile
@@ -1,7 +1,7 @@
 # Used for image metadata only
 # Describes the version of the Dockerfile, *not* the version of the bundled Hazelcast binary as this is/can be controlled externally
 # Dockerfile needs some concept of versioning so that the release pipeline can tag/archive with an appropriate label
-ARG HZ_VERSION=5.8.0-SNAPSHOT
+ARG HZ_VERSION=6.0.0-SNAPSHOT
 
 # Build constants
 ARG HZ_HOME="/opt/hazelcast"
--- a/hazelcast-oss/Dockerfile
+++ b/hazelcast-oss/Dockerfile
@@ -1,7 +1,7 @@
 # Used for image metadata only
 # Describes the version of the Dockerfile, *not* the version of the bundled Hazelcast binary as this is/can be controlled externally
 # Dockerfile needs some concept of versioning so that the release pipeline can tag/archive with an appropriate label
-ARG HZ_VERSION=5.8.0-SNAPSHOT
+ARG HZ_VERSION=6.0.0-SNAPSHOT
 
 # Build constants
 ARG HZ_HOME="/opt/hazelcast"


JackPGreen (changes requested):
I don’t understand why we are updating manually - that seems more error prone, easier to miss some obscure step.

nishaatr:
> I don’t understand why we are updating manually - that seems more error prone, easier to miss some obscure step.

This is what pipeline does. There is nothing else I believe
No different to https://github.com/hazelcast/hazelcast-docker/commit/0cf75e6dbe417d0c237a4f05398acd0fcac7ad48

ps: this way avoids creation and deletion of `5.8.0/5.8.z` branches.

JackPGreen:
> > I don’t understand why we are updating manually - that seems more error prone, easier to miss some obscure step.
> 
> This is what pipeline does. There is nothing else I believe No different to [0cf75e6](https://github.com/hazelcast/hazelcast-docker/commit/0cf75e6dbe417d0c237a4f05398acd0fcac7ad48)
> 
> ps: this way avoids creation and deletion of `5.8.0/5.8.z` branches.

My concern isn’t if this change is right, it’s that we might miss a whole area of changes. EG schemas (I know not required but an example). The previous down-versioning was manual, tedious and non-exhaustive - we were mopping up changes afterwards.

JamesHazelcast:
> My concern isn’t if this change is right, it’s that we might miss a whole area of changes. EG schemas (I know not required but an example). The previous down-versioning was manual, tedious and non-exhaustive - we were mopping up changes afterwards.

It LGTM based on previous development cycle bumps. I appreciate what you're saying, we want to avoid even manual verification, but in this case we've already done those steps (rightly or wrongly), so I don't see the benefit in blocking this PR - unless there's something specific you've spotted or are concerned about?

nishaatr:
> > My concern isn’t if this change is right, it’s that we might miss a whole area of changes. EG schemas (I know not required but an example). The previous down-versioning was manual, tedious and non-exhaustive - we were mopping up changes afterwards.
> 
> It LGTM based on previous development cycle bumps. I appreciate what you're saying, we want to avoid even manual verification, but in this case we've already done those steps (rightly or wrongly), so I don't see the benefit in blocking this PR - unless there's something specific you've spotted or are concerned about?

Just to reiterate, doing it via Pipeline means:
1. Revert https://github.com/hazelcast/hazelcast-mono/pull/6982
2. Run [release-branch-automation](https://jenkins.hazelcast.com/view/DevInfra/job/release-automation/job/release-branch-automation/)
3. Delete `5.8.0` and `5.8.z` in repos
4. Delete `*-5.8.xsd`
5. Do [Add new version to Versions.java](https://hazelcast.atlassian.net/wiki/spaces/DI/pages/4681138242/Create+branch+5.x#Add-new-version-to-Versions.java)

Doing it via PR:
1. take @JamesHazelcast changes. Had cursory look and looks ok. @JamesHazelcast 
2. Need this PR and https://github.com/hazelcast/hazelcast-packaging/pull/365 

As for docs changes, there is no other option but manual updates (as no release)

@JackPGreen totally understand your sentiments and I would have preferred this went via pipeline. But form where I am standing, PR is the simplest option in this case

Can you recall what specifically you had to mop?

JackPGreen:
> Can you recall what specifically you had to mop?

Unfortunately, no.

JamesHazelcast (approved):

nishaatr:
Build is taking way too long for RHEL
Its waiting on RedHat infra at the moment
Build now has progressed beyond earlier HZ download failure 
So merging as should be fine!

PR #1300, opened 2026-08-13
In smoke test, merge Docker logs into main job execution logs instead of distinct artifact

JackPGreen:
When we run the Docker container as part of the smoke test, we run it in [detached mode](https://docs.docker.com/reference/cli/docker/container/run/#detach) to allow subsequent steps to continue. But when running in a detached way, logs are not displayed - so we have a separate step to dump and upload those logs as an artifact.

GitHub artifacts [consume storage credits](https://docs.github.com/en/billing/concepts/product-billing/github-actions#how-storage-billing-works), and this artifact exists only as a workaround to capture the logs.

[Using the new `background` functionality](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepsbackground), we can then convert this detached execution into an _attached_ execution of a background step - and the logs are then rendered as part of the job logs, and no storage credits are consumed:
image

### DIFF
--- a/.github/workflows/ee-nlc-tag-package.yml
+++ b/.github/workflows/ee-nlc-tag-package.yml
@@ -143,7 +143,7 @@ jobs:
           docker run \
             --name "${TEST_CONTAINER_NAME}" \
             --env HZ_INSTANCETRACKING_FILENAME \
-            --expose 5701:5701 \
+            -p5701:5701 \
             "${LOCAL_IMAGE_NAME}:${{ steps.get-tags-to-push.outputs.primary-tag }}"
         env:
           HZ_INSTANCETRACKING_FILENAME: instance-tracking.txt
--- a/.github/workflows/ee-nlc-tag-package.yml
+++ b/.github/workflows/ee-nlc-tag-package.yml
@@ -142,8 +142,8 @@ jobs:
         run: |
           docker run \
             --name "${TEST_CONTAINER_NAME}" \
-            -e HZ_INSTANCETRACKING_FILENAME \
-            -p5701:5701 \
+            --env HZ_INSTANCETRACKING_FILENAME \
+            --expose 5701:5701 \
             "${LOCAL_IMAGE_NAME}:${{ steps.get-tags-to-push.outputs.primary-tag }}"
         env:
           HZ_INSTANCETRACKING_FILENAME: instance-tracking.txt
--- a/.github/workflows/ee-nlc-tag-package.yml
+++ b/.github/workflows/ee-nlc-tag-package.yml
@@ -136,6 +136,18 @@ jobs:
           # The distribution is large and no longer required now the image has been built
           rm "${{ steps.download-hz-dist.outputs.output_file }}"
 
+      - name: Run test container
+        id: test-container
+        background: true
+        run: |
+          docker run \
+            --name "${TEST_CONTAINER_NAME}" \
+            -e HZ_INSTANCETRACKING_FILENAME \
+            -p5701:5701 \
+            "${LOCAL_IMAGE_NAME}:${{ steps.get-tags-to-push.outputs.primary-tag }}"
+        env:
+          HZ_INSTANCETRACKING_FILENAME: instance-tracking.txt
+
       - name: Run smoke test against image
         timeout-minutes: 2
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/ee-nlc-tag-package.yml
```diff
@@ -136,6 +136,18 @@ jobs:
...
 
+      - name: Run test container
+        id: test-container
+        background: true
+        run: |
+          docker run \
+            --name "${TEST_CONTAINER_NAME}" \
+            -e HZ_INSTANCETRACKING_FILENAME \
+            -p5701:5701 \
+            "${LOCAL_IMAGE_NAME}:${{ steps.get-tags-to-push.outputs.primary-tag }}"
```
nishaatr, line 147: use full option names for all docker calls
```
docker run \
    --name "${TEST_CONTAINER_NAME}" \
    --env HZ_INSTANCETRACKING_FILENAME \
    --publish 5701:5701 \
    "${LOCAL_IMAGE_NAME}:${{ steps.get-tags-to-push.outputs.primary-tag }}"

```

.github/workflows/test_published_images.yml
```diff
@@ -171,7 +171,11 @@ jobs:
             fi
 
             # Compute tag by concatenating tag_elements
-            .github/scripts/simple-smoke-test.sh "${repository}/${image_name}":$(IFS=- ; echo "${tag_elements[*]}") "${CONTAINER_NAME}" "${{ matrix.distribution-type }}" "${{ inputs.EXPECTED_HZ_VERSION }}" "${expected_jdk_version}"
+            local image="${repository}/${image_name}":$(IFS=- ; echo "${tag_elements[*]}")
+
+            docker run -it --name "${CONTAINER_NAME}" -e HZ_LICENSEKEY -e HZ_INSTANCETRACKING_FILENAME -d -p5701:5701 "${image}"
```
nishaatr: multiline + full option names

.github/workflows/test_published_images.yml
```diff
@@ -171,7 +171,11 @@ jobs:
...
+            .github/scripts/simple-smoke-test.sh "${image}" "${CONTAINER_NAME}" "${{ matrix.distribution-type }}" "${{ inputs.EXPECTED_HZ_VERSION }}" "${expected_jdk_version}"
```
nishaatr: existing but worth multilining!

.github/workflows/test_published_images.yml
nishaatr: use `background:` ?

.github/workflows/ee-nlc-tag-package.yml
```diff
@@ -136,6 +136,18 @@ jobs:
           # The distribution is large and no longer required now the image has been built
           rm "${{ steps.download-hz-dist.outputs.output_file }}"
 
+      - name: Run test container
+        id: test-container
+        background: true
+        run: |
```
nishaatr, line 142: just checking no need for `remove_container_if_exists()` because running as `background`?

nishaatr (changes requested):

.github/workflows/test_published_images.yml
JackPGreen: We can’t as this job executes the test with multiple different containers in a single execution.
It probably could be refactored to be a matrix or something but would be non-trivial, and it’s only used infrequently for post release verification.

.github/workflows/ee-nlc-tag-package.yml
JackPGreen: This was required to support local execution so that you could just run the script without any setup / tear down. Now it’s more tightly coupled to GitHub Actions and doesn’t start the containers, it’s not required. When executed in (ephemeral) CI runners, we’d never expect the container to already exist.

nishaatr: makes sense

nishaatr (approved):
Approving as remaining 2 issues are nits.
Sonar needs attention!

.github/workflows/ee-nlc-tag-package.yml
JackPGreen: f564fa5a836901678471995a04bdeafa589e4e78

.github/workflows/test_published_images.yml
JackPGreen: ae1a9480c53727edd2268e4552fce99f20db11a8

.github/workflows/test_published_images.yml
JackPGreen: 32e0c252f684102fe153960d3d20cc23c70cda0f

PR #1301, opened 2026-08-15
Sign Docker images with `cosign`

JackPGreen:
See:
- https://docs.docker.com/dhi/explore/security-concepts/signatures
- https://edu.chainguard.dev/open-source/sigstore/cosign/an-introduction-to-cosign/

We can use `cosign` in conjunction with our GitHub OIDC token to sign that the image deployed to Docker registries was built inside an action running within our GitHub organisation - as a pose to by some third party actor pushing _their_ build to the registries in some supply chain attack.

Docker's "hardened images" are signed.

Examples of usage:
```
% cosign verify {sandbox-hazelcast-enterprise-image} --certificate-identity dummy-to-query --certificate-oidc-issuer https://token.actions.githubusercontent.com

{...} expected SAN value "dummy-to-query", got "https://github.com/hazelcast/hazelcast-docker/.github/workflows/tag_image_package.yml@refs/pull/1301/merge"
```

```
% cosign verify {sandbox-hazelcast-enterprise-image}  --certificate-identity https://github.com/hazelcast/hazelcast-docker/.github/workflows/tag_image_package.yml@refs/pull/1301/merge --certificate-oidc-issuer https://token.actions.githubusercontent.com


Verification for index.docker.io/{sandbox-hazelcast-enterprise-image} --
The following checks were performed on each of these signatures:
  - The cosign claims were validated
  - Existence of the claims in the transparency log was verified offline
  - The code-signing certificate was verified using trusted certificate authority certificates

[{"critical":{"identity":{"docker-reference":"index.docker.io/{sandbox-hazelcast-enterprise-image}"},"image":{"docker-manifest-digest":"sha256:2282ba9f8f224b9c86209f36e780563651d2eba64bd8bd24d17e8ab90e2bf87e"},"type":"https://sigstore.dev/cosign/sign/v1"},"optional":{}}]
```

Changes:
- sign images when building
- migrate from `skopeo` to [`oras`](https://oras.land/docs/commands/oras_cp) for copying images
   - `skopeo` only copies _images_
   - `oras` copies OCI artifacts, which includes images _and attached metadata_ (e.g. the signatures)

### DIFF
--- a/.github/workflows/ee-nlc-tag-package.yml
+++ b/.github/workflows/ee-nlc-tag-package.yml
@@ -117,6 +117,7 @@ jobs:
           HZ_REVISION: ${{ inputs.HZ_REVISION }}
 
       - name: Build and push image
+        id: build-and-push
         uses: docker/build-push-action@v7
         env:
           # Fixed timestamp (0 = Unix Epoch) needed for reproducible builds
@@ -129,6 +130,11 @@ jobs:
           tags: ${{ env.LOCAL_IMAGE_NAME}}:${{ steps.get-tags-to-push.outputs.primary-tag }}
           build-args: |
             JDK_VERSION=${{ matrix.jdk }}
+
+      - uses: ./.github/actions/sign-image
+        with:
+          image-name: ${{ env.LOCAL_IMAGE_NAME }}
+          digest: ${{ steps.build-and-push.outputs.digest }}
 
       - name: Delete the distribution file
         run: |
@@ -185,15 +191,16 @@ jobs:
           username: ${{ env.NLC_IMAGE_REGISTRY_USERNAME }}
           password: ${{ env.NLC_IMAGE_REGISTRY_TOKEN }}
 
+      - uses: oras-project/setup-oras@1d808f7d7f6995cc68b7bf507bfe5c5446e1dc9d # v2.0.1
+
       - name: Push image to remote registry
         run: |
-          # Skopeo expects HTTPS registry but local is HTTP
-          skopeo copy \
-            --all \
-            --retry-times 10 \
-            --src-tls-verify=false \
-            docker://${LOCAL_IMAGE_NAME}:${{ steps.get-tags-to-push.outputs.primary-tag }} \
-            docker://${NLC_IMAGE_REGISTRY_URL}/hazelcast_cloud/${NLC_IMAGE_NAME}-${{ vars.PREPROD_REGISTRY }}:${{ steps.get-tags-to-push.outputs.primary-tag }}
+          # oras expects HTTPS registry but local is HTTP
+          oras cp \
+            --from-plain-http \
+            --recursive \
+            ${LOCAL_IMAGE_NAME}:${{ steps.get-tags-to-push.outputs.primary-tag }} \
+            ${NLC_IMAGE_REGISTRY_URL}/hazelcast_cloud/${NLC_IMAGE_NAME}-${{ vars.PREPROD_REGISTRY }}:${{ steps.get-tags-to-push.outputs.primary-tag }}
         env:
           NLC_IMAGE_NAME: ${{ vars.NLC_IMAGE_NAME }}
 
--- a/.github/workflows/tag_image_package.yml
+++ b/.github/workflows/tag_image_package.yml
@@ -126,6 +126,7 @@ jobs:
           HZ_REVISION: ${{ inputs.HZ_REVISION }}
 
       - name: Build and push image
+        id: build-and-push
         uses: docker/build-push-action@v7
         env:
           # Fixed timestamp (like 0 EPOCH) is needed for reproducible builds
@@ -138,6 +139,11 @@ jobs:
           tags: ${{ env.LOCAL_IMAGE_NAME }}:${{ steps.get-tags-to-push.outputs.primary-tag }}
           build-args: |
             JDK_VERSION=${{ matrix.jdk }}
+
+      - uses: ./.github/actions/sign-image
+        with:
+          image-name: ${{ env.LOCAL_IMAGE_NAME }}
+          digest: ${{ steps.build-and-push.outputs.digest }}
 
       - name: Delete the distribution file
         run: |
@@ -181,15 +187,16 @@ jobs:
         with:
           username: ${{ vars.DOCKERHUB_ORGANIZATION }}
 
+      - uses: oras-project/setup-oras@1d808f7d7f6995cc68b7bf507bfe5c5446e1dc9d # v2.0.1
+
       - name: Push image to remote registry
         run: |
-          # Skopeo expects HTTPS registry but local is HTTP
-          skopeo copy \
-            --all \
-            --retry-times 10 \
-            --src-tls-verify=false \
-            docker://${LOCAL_IMAGE_NAME}:${{ steps.get-tags-to-push.outputs.primary-tag }} \
-            docker://${{ vars.DOCKERHUB_ORGANIZATION }}/${{ vars.PREPROD_REGISTRY }}-${{ matrix.distribution-type.image-name }}:${{ steps.get-tags-to-push.outputs.primary-tag }}
+          # oras expects HTTPS registry but local is HTTP
+          oras cp \
+            --from-plain-http \
+            --recursive \
+            ${LOCAL_IMAGE_NAME}:${{ steps.get-tags-to-push.outputs.primary-tag }} \
+            docker.io/${{ vars.DOCKERHUB_ORGANIZATION }}/${{ vars.PREPROD_REGISTRY }}-${{ matrix.distribution-type.image-name }}:${{ steps.get-tags-to-push.outputs.primary-tag }}
 
   failure-notifications:
     name: Failure notification
--- a/.github/workflows/tag_image_promote_docker_registry.yml
+++ b/.github/workflows/tag_image_promote_docker_registry.yml
@@ -99,22 +99,20 @@ jobs:
           username: ${{ steps.authenticate-jfrog-write-dev-account.outputs.user }}
           password: ${{ steps.authenticate-jfrog-write-dev-account.outputs.token }}
 
+      - uses: oras-project/setup-oras@1d808f7d7f6995cc68b7bf507bfe5c5446e1dc9d # v2.0.1
+
       - name: Push image to remote registry
         run: |
           if [[ "${{ matrix.distribution-type.label }}" == "oss" && "${HZ_VERSION}" =~ SNAPSHOT ]]; then
             remote_registry=docker.hazelcast.com/docker/hazelcast
           else
-            remote_registry=${{ vars.DOCKERHUB_ORGANIZATION }}
+            remote_registry=docker.io/${{ vars.DOCKERHUB_ORGANIZATION }}
           fi
 
-          for tag in ${TAGS_TO_PUSH}
-          do
-            skopeo copy \
-              --all \
-              --retry-times 10 \
-              docker://${{ vars.DOCKERHUB_ORGANIZATION }}/${{ vars.PREPROD_REGISTRY }}-${{ matrix.distribution-type.image-name }}:${{ steps.get-tags-to-push.outputs.primary-tag }} \
-              docker://${remote_registry}/${{ matrix.distribution-type.image-name }}:${tag}
-          done
+          oras cp \
+            --recursive \
+            docker.io/${{ vars.DOCKERHUB_ORGANIZATION }}/${{ vars.PREPROD_REGISTRY }}-${{ matrix.distribution-type.image-name }}:${{ steps.get-tags-to-push.outputs.primary-tag }} \
+            ${remote_registry}/${{ matrix.distribution-type.image-name }}:${TAGS_TO_PUSH// /,}
         env:
           HZ_VERSION: ${{ inputs.HZ_VERSION }}
           TAGS_TO_PUSH: ${{ steps.get-tags-to-push.outputs.all-destination-tags }}
--- a/.github/workflows/tag_image_promote_rhel.yml
+++ b/.github/workflows/tag_image_promote_rhel.yml
@@ -162,16 +162,18 @@ jobs:
           # Add date suffix to create unique output tag (e.g. 5.4.1-jdk17-20250909130347)
           echo "tag=${{ steps.get-tags-to-push.outputs.primary-tag }}-$(date +"%Y%m%d%H%M%S")" >> ${GITHUB_OUTPUT}
 
+      - if: steps.image-already-published.outputs.IMAGE_ALREADY_PUBLISHED != 'true'   
+        uses: oras-project/setup-oras@1d808f7d7f6995cc68b7bf507bfe5c5446e1dc9d # v2.0.1
+
       - name: Copy new unique tag to RedHat Container Registry
         if: steps.image-already-published.outputs.IMAGE_ALREADY_PUBLISHED != 'true'
         run: |
           # Deliberately only copy a single architecture - https://github.com/hazelcast/hazelcast-docker/pull/1042#issuecomment-3141395993
-          skopeo copy \
-            --retry-times 10 \
-            --override-os "${PLATFORM_OS}" \
-            --override-arch "${PLATFORM_ARCH}" \
-            docker://${{ vars.DOCKERHUB_ORGANIZATION }}/${{ vars.PREPROD_REGISTRY }}-${{ vars.DOCKERHUB_EE_IMAGE_NAME }}:${{ steps.get-tags-to-push.outputs.primary-tag }} \
-            docker://${{ steps.scan_registry_secrets.outputs.SCAN_REPOSITORY }}:${{ steps.unique-destination-tag.outputs.tag }}
+          oras cp \
+            --recursive \
+            --platform ${PLATFORM_OS}/${PLATFORM_ARCH} \
+            docker.io/${{ vars.DOCKERHUB_ORGANIZATION }}/${{ vars.PREPROD_REGISTRY }}-${{ vars.DOCKERHUB_EE_IMAGE_NAME }}:${{ steps.get-tags-to-push.outputs.primary-tag }} \
+            docker.io/${{ steps.scan_registry_secrets.outputs.SCAN_REPOSITORY }}:${{ steps.unique-destination-tag.outputs.tag }}
 
       - name: Install `preflight` OpenShift tool from GitHub
         if: steps.image-already-published.outputs.IMAGE_ALREADY_PUBLISHED != 'true'
@@ -227,19 +229,13 @@ jobs:
       - name: Copy remaining tags to RedHat Container Registry
         if: steps.image-already-published.outputs.IMAGE_ALREADY_PUBLISHED != 'true'
         run: |
-          for tag in ${TAGS_TO_PUSH}
-          do
-            echo "Copying tag: ${FULL_PRIMARY_TAG} -> ${tag}"
-            skopeo copy \
-              --retry-times 10 \
-              --override-os "${PLATFORM_OS}" \
-              --override-arch "${PLATFORM_ARCH}" \
-              docker://${FULL_PRIMARY_TAG} \
-              docker://${{ steps.scan_registry_secrets.outputs.SCAN_REPOSITORY }}:${tag}
-          done
+          oras cp \
+            --recursive \
+            --platform ${PLATFORM_OS}/${PLATFORM_ARCH} \
+            docker.io/${{ vars.DOCKERHUB_ORGANIZATION }}/${{ vars.PREPROD_REGISTRY }}-${{ vars.DOCKERHUB_EE_IMAGE_NAME }}:${{ steps.get-tags-to-push.outputs.primary-tag }} \
+            ${{ steps.scan_registry_secrets.outputs.SCAN_REPOSITORY }}:${TAGS_TO_PUSH// /,}
         env:
           TAGS_TO_PUSH: ${{ steps.get-tags-to-push.outputs.all-destination-tags }}
-          FULL_PRIMARY_TAG: ${{ vars.DOCKERHUB_ORGANIZATION }}/${{ vars.PREPROD_REGISTRY }}-${{ vars.DOCKERHUB_EE_IMAGE_NAME }}:${{ steps.get-tags-to-push.outputs.primary-tag }}
 
       - name: Check tags published
         # Just because the tag has been pushed to the registry, does not guarantee it's accessible
--- a/.github/actions/sign-image/action.yml
+++ b/.github/actions/sign-image/action.yml
@@ -1,0 +1,19 @@
+name: Sign image with Cosign
+inputs:
+  image-name:
+    required: true
+  digest:
+    required: true
+runs:
+  using: "composite"
+  steps:
+    - uses: sigstore/cosign-installer@6f9f17788090df1f26f669e9d70d6ae9567deba6 # v4.1.2
+
+      # https://github.com/sigstore/cosign-installer#usage
+    - name: Sign the image
+      shell: bash
+      run: |
+        cosign sign --yes "${IMAGE_NAME}@${DIGEST}"
+      env:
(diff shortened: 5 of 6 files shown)


JackPGreen:
@nishaatr @ldziedziul are you able to review, please?

.github/workflows/ee-nlc-tag-package.yml
```diff
@@ -185,15 +191,16 @@ jobs:
...
         run: |
-          # Skopeo expects HTTPS registry but local is HTTP
-          skopeo copy \
-            --all \
-            --retry-times 10 \
-            --src-tls-verify=false \
-            docker://${LOCAL_IMAGE_NAME}:${{ steps.get-tags-to-push.outputs.primary-tag }} \
-            docker://${NLC_IMAGE_REGISTRY_URL}/hazelcast_cloud/${NLC_IMAGE_NAME}-${{ vars.PREPROD_REGISTRY }}:${{ steps.get-tags-to-push.outputs.primary-tag }}
+          # oras expects HTTPS registry but local is HTTP
+          oras cp \
```
nishaatr, line 200: may be make local action to:
1. oras-project/setup-oras
2. oras cp

In similar vein as `sign-image`

.github/workflows/tag_image_promote_rhel.yml
```diff
@@ -227,19 +229,13 @@ jobs:
...
-          do
-            echo "Copying tag: ${FULL_PRIMARY_TAG} -> ${tag}"
-            skopeo copy \
-              --retry-times 10 \
-              --override-os "${PLATFORM_OS}" \
-              --override-arch "${PLATFORM_ARCH}" \
-              docker://${FULL_PRIMARY_TAG} \
-              docker://${{ steps.scan_registry_secrets.outputs.SCAN_REPOSITORY }}:${tag}
-          done
+          oras cp \
```
nishaatr, line 238: Just checking `cosign verify ...`  works with published RHEL image ?

.github/workflows/tag_image_promote_docker_registry.yml
```diff
@@ -99,22 +99,20 @@ jobs:
...
 
+      - uses: oras-project/setup-oras@1d808f7d7f6995cc68b7bf507bfe5c5446e1dc9d # v2.0.1
+
       - name: Push image to remote registry
         run: |
           if [[ "${{ matrix.distribution-type.label }}" == "oss" && "${HZ_VERSION}" =~ SNAPSHOT ]]; then
             remote_registry=docker.hazelcast.com/docker/hazelcast
           else
-            remote_registry=${{ vars.DOCKERHUB_ORGANIZATION }}
+            remote_registry=docker.io/${{ vars.DOCKERHUB_ORGANIZATION }}
```
nishaatr: why this change?

.github/workflows/tag_image_promote_rhel.yml
```diff
@@ -162,16 +162,18 @@ jobs:
...
           # Deliberately only copy a single architecture - https://github.com/hazelcast/hazelcast-docker/pull/1042#issuecomment-3141395993
-          skopeo copy \
-            --retry-times 10 \
-            --override-os "${PLATFORM_OS}" \
-            --override-arch "${PLATFORM_ARCH}" \
-            docker://${{ vars.DOCKERHUB_ORGANIZATION }}/${{ vars.PREPROD_REGISTRY }}-${{ vars.DOCKERHUB_EE_IMAGE_NAME }}:${{ steps.get-tags-to-push.outputs.primary-tag }} \
-            docker://${{ steps.scan_registry_secrets.outputs.SCAN_REPOSITORY }}:${{ steps.unique-destination-tag.outputs.tag }}
+          oras cp \
+            --recursive \
+            --platform ${PLATFORM_OS}/${PLATFORM_ARCH} \
```
nishaatr, line 174: Just checking that all `--platforms` are indeed published?
Likely does but thought I check as migrating from `skopeo`!

nishaatr (changes requested):

.github/workflows/tag_image_promote_docker_registry.yml
JackPGreen: Skopeo/Docker implicitly use Dockerhub if no registry specified - `oras` does not and requires it to be specified.

.github/workflows/tag_image_promote_rhel.yml
JackPGreen: We only publish one platform.

.github/workflows/ee-nlc-tag-package.yml
JackPGreen: Maybe a whole composite action for 2 steps is overengineered - i.e. remove `sign-image`?

The `oras cp` executions are _subtley_ different so I'm not sure centralising would help.

.github/workflows/tag_image_promote_rhel.yml
JackPGreen: Good question, no idea and struggling to test in sandbox to check. But it should do - and if it doesn’t, it’s a repo-thing outside of our control.

.github/workflows/ee-nlc-tag-package.yml
nishaatr: > Maybe a whole composite action for 2 steps is overengineered - i.e. remove sign-image?

I think yes. I suspect you did action to have the version in a single place, so I thought do the same for `cp` and add option for `--from-plain-http` and `--platform` (may be this could be defaulted). But thats making it complicated so best to inline everything i.e. remove `sign-image`.

.github/workflows/tag_image_promote_docker_registry.yml
nishaatr: ok

.github/workflows/tag_image_promote_rhel.yml
nishaatr: Actually, I meant for where we publish multiple platforms for RHEL? 

Also, before we had `skopeo --all` but like [here ](https://github.com/hazelcast/hazelcast-docker/pull/1301/changes#diff-591e4461999f32ca2b0e49e348908138c508d91831299cf847c056e92be19daaR195) I dont see mirroring of that i.e. no `--platform`. Why is that?

.github/workflows/tag_image_promote_rhel.yml
nishaatr: Whats the issue?
I thought we have `sandbox` repo in RHEL registry so should be able to publish and test

.github/workflows/tag_image_promote_rhel.yml
JackPGreen: > Actually, I meant for where we publish multiple platforms for RHEL?

We only publish a single platform for RedHat:
https://github.com/hazelcast/hazelcast-docker/blob/790da99857580f446272e0fa516c40abdb9389da/.github/workflows/tag_image_promote_rhel.yml#L165-L174

> Also, before we had `skopeo --all` but like [here ](https://github.com/hazelcast/hazelcast-docker/pull/1301/changes#diff-591e4461999f32ca2b0e49e348908138c508d91831299cf847c056e92be19daaR195) I dont see mirroring of that i.e. no `--platform`. Why is that?

By default, `skopeo` copies _current_ (runtime) platform only. So for our purposes, either we override with `--all` to copy all platforms, or we explicitly specify a platform (which may or may not be what were actually running on).

`oras` copies all platforms by default - so we don't need to specify. But we can specify explicit platform if required. [Related docs](http://github.com/oras-project/oras/issues/1945).

nishaatr: ok. kind of on the fence, but i think it will be clearer if we pass `--platform` - in part for consistency

nishaatr: btw...any testing done to pull (and run) the image or even check if it appears in the UI?
You had issues with `verify` https://github.com/hazelcast/hazelcast-docker/pull/1301#discussion_r3913836466 so just checking

JackPGreen: > ok. kind of on the fence, but i think it will be clearer if we pass `--platform` - in part for consistency

Where do you mean? We can't replace `--all` with `--platform`, they aren't equivalent.

.github/workflows/ee-nlc-tag-package.yml
JackPGreen: b97ced0fe2cc3cd317f775ab7951e637134d4e15

.github/workflows/tag_image_promote_rhel.yml
nishaatr: i meant we use `--platform` [here](https://github.com/hazelcast/hazelcast-docker/pull/1301/changes#diff-bb599021c1119b611a6433ffdb48d3738307faa39f02761d3f73838f8f75ac81R234)

and do the same [here ](https://github.com/hazelcast/hazelcast-docker/pull/1301/changes#diff-f23c04297854b0de26c83e33985d9ee8bbd18d70dca9291501fa34d7ec3b32cdR201) and [here ](https://github.com/hazelcast/hazelcast-docker/pull/1301/changes#diff-fb538f40c1fa1ee4a7c8d4c724028da0433f997ad8e3312476e40d9a99aacb73R112). So basically use `--platform` everywhere

But its fine as is as. its clearer to me now!

.github/workflows/tag_image_promote_rhel.yml
JackPGreen: > Good question, no idea and struggling to test in sandbox to check. But it should do - and if it doesn’t, it’s a repo-thing outside of our control.



> Whats the issue? I thought we have `sandbox` repo in RHEL registry so should be able to publish and test

Although we publish to sandbox during PR, RedHat doesn't allow identical image (re)publishing. And as the images are reproducible, it means the PR builder doesn't always publish to RedHat.

I was able to spot an issue with publish to RedHat though and fix in https://github.com/hazelcast/hazelcast-docker/pull/1301/commits/e1dc72c59ffbbad7da3195c977139d653cb2dc92

I've also added tests to check the signatures exist in source/target repos to catch such issues in future https://github.com/hazelcast/hazelcast-docker/pull/1301/changes/b63b22d4bf152aab6ea5f258875afb6804db3c2f

nishaatr: thats good. 
I thought we have a force rebuild option if I recall?
Otherwise, how does the rebuild work?
Just asking for my ref!

JackPGreen: The force build is used when checking if images are outdated / need a rebuild, and skips the check and builds new images regardless. But we have the same “issue” there that due to reproducible builds, the resultant image will still be the same and RedHat would fail - hence why we always check first in RedHat and skip if already there.

nishaatr: noted
could you not unpublish/delete any existing `sandbox` images and deploy/test that?

JackPGreen: Right, was a bit of a faff to unpublish/delete the images via the RedHat UI but I've done it and the tests passed.

I've also had to add some polling to make it stable in CI.

.github/workflows/ee-nlc-tag-package.yml
```diff
@@ -210,12 +210,19 @@ jobs:
...
+        timeout-minutes: 1
         run: |
+          source /dev/stdin <<< "$(curl --fail --retry 5 --retry-all-errors --show-error --silent https://raw.githubusercontent.com/hazelcast/github-actions-common-scripts/main/logging.functions.sh)"
+
           # https://github.com/sigstore/cosign-installer#usage
-          cosign verify \
+          # Image signatures can take a couple of seconds to appear after the main image (suspect "oras cp --recursive" is not fully atomic), so poll for completion
+          while ! cosign verify \
             --certificate-identity-regexp "${GITHUB_SERVER_URL//./\\.}/${GITHUB_REPOSITORY}/.*" \
             --certificate-oidc-issuer https://token.actions.githubusercontent.com \
```
nishaatr: I believe you can use ACTIONS_ID_TOKEN_REQUEST_URL instead of https://token.actions.githubusercontent.com so its not hardcoded

Apply everywhere!

nishaatr (changes requested):

nishaatr:
I think 1 sec may be too quick for sleep?
Thinking of RHEL as we know can be slow

.github/workflows/ee-nlc-tag-package.yml
JackPGreen: > I believe you can use ACTIONS_ID_TOKEN_REQUEST_URL instead of https://token.actions.githubusercontent.com so its not hardcoded
> 
> Apply everywhere!

[`ACTIONS_ID_TOKEN_REQUEST_URL` evaluates to something completely different (`https://run-actions-1-azure-eastus.actions.githubusercontent.com/207//idtoken/48411e6f-747a-4124-8a33-ae8c5f1501dd/544778bb-56ae-541c-b014-016d23fa8e45?api-version=2.0`). Are you sure?](https://github.com/JackPGreen/backport-test/actions/runs/33904934847/job/101127529217#step:2:97)

JackPGreen:
> I think 1 sec may be too quick for sleep? Thinking of RHEL as we know can be slow

It's polling for completion every second - not waiting one second only.

.github/workflows/tag_image_package.yml
```diff
@@ -181,15 +192,33 @@ jobs:
...
+          # oras expects HTTPS registry but local is HTTP
+          oras cp \
+            --from-plain-http \
+            --recursive \
+            ${LOCAL_IMAGE_NAME}:${{ steps.get-tags-to-push.outputs.primary-tag }} \
+            docker.io/${{ vars.DOCKERHUB_ORGANIZATION }}/${{ vars.PREPROD_REGISTRY }}-${{ matrix.distribution-type.image-name }}:${{ steps.get-tags-to-push.outputs.primary-tag }}
+
+      - uses: sigstore/cosign-installer@6f9f17788090df1f26f669e9d70d6ae9567deba6 # v4.1.2
+
+      - name: Validate image signature
```
ldziedziul: I know the composite action was dropped in https://github.com/hazelcast/hazelcast-docker/pull/1301#discussion_r3913702134, but this verify step (cosign-installer + logging curl + poll loop) is now the same ~16 lines in 5 workflows, only the image ref changes. Also the retry/timeout tuning happens in one place. WDYT?

.github/workflows/tag_image_package.yml
```diff
@@ -181,15 +192,33 @@ jobs:
...
+        timeout-minutes: 1
+        run: |
+          source /dev/stdin <<< "$(curl --fail --retry 5 --retry-all-errors --show-error --silent https://raw.githubusercontent.com/hazelcast/github-actions-common-scripts/main/logging.functions.sh)"
+
+          # https://github.com/sigstore/cosign-installer#usage
+          # Image signatures can take a couple of seconds to appear after the main image (suspect "oras cp --recursive" is not fully atomic), so poll for completion
+          while ! cosign verify \
```
ldziedziul: This loops on *any* failure (no signature, identity mismatch, registry without referrers support...) until `timeout-minutes` kills it, so the job fails as "timed out" instead of a verify error, and `echodebug` is invisible unless debug logging is on. Couldn't we bound it and fail explicitly? E.g.

```bash
MAX_ATTEMPTS=30
verified=false

for ((attempt = 1; attempt <= MAX_ATTEMPTS; attempt++)); do
  if cosign verify ... "${IMAGE}"; then
    verified=true
    break
  fi

  if (( attempt < MAX_ATTEMPTS )); then
    echonotice "Signature verification failed - additional annotations may not yet be available (attempt ${attempt}/${MAX_ATTEMPTS}), retrying..."
    sleep 1
  fi
done

if [[ "$verified" != true ]]; then
  echoerr "Could not verify the signature of ${IMAGE} after ${MAX_ATTEMPTS} attempts"
  exit 1
fi
```

.github/workflows/tag_image_package.yml
```diff
@@ -181,15 +192,33 @@ jobs:
...
+            --recursive \
+            ${LOCAL_IMAGE_NAME}:${{ steps.get-tags-to-push.outputs.primary-tag }} \
+            docker.io/${{ vars.DOCKERHUB_ORGANIZATION }}/${{ vars.PREPROD_REGISTRY }}-${{ matrix.distribution-type.image-name }}:${{ steps.get-tags-to-push.outputs.primary-tag }}
+
+      - uses: sigstore/cosign-installer@6f9f17788090df1f26f669e9d70d6ae9567deba6 # v4.1.2
+
+      - name: Validate image signature
+        timeout-minutes: 1
+        run: |
+          source /dev/stdin <<< "$(curl --fail --retry 5 --retry-all-errors --show-error --silent https://raw.githubusercontent.com/hazelcast/github-actions-common-scripts/main/logging.functions.sh)"
```
ldziedziul: `. .github/scripts/logging.functions.sh` like the other workflows - the wrapper does exactly this curl (e.g. `tag_image_promote_rhel.yml` does it that way a few steps below the new one). Same in the other 4 copies.

.github/workflows/tag_image_package.yml
```diff
@@ -181,15 +192,33 @@ jobs:
...
-            docker://${LOCAL_IMAGE_NAME}:${{ steps.get-tags-to-push.outputs.primary-tag }} \
-            docker://${{ vars.DOCKERHUB_ORGANIZATION }}/${{ vars.PREPROD_REGISTRY }}-${{ matrix.distribution-type.image-name }}:${{ steps.get-tags-to-push.outputs.primary-tag }}
+          # oras expects HTTPS registry but local is HTTP
+          oras cp \
+            --from-plain-http \
+            --recursive \
+            ${LOCAL_IMAGE_NAME}:${{ steps.get-tags-to-push.outputs.primary-tag }} \
+            docker.io/${{ vars.DOCKERHUB_ORGANIZATION }}/${{ vars.PREPROD_REGISTRY }}-${{ matrix.distribution-type.image-name }}:${{ steps.get-tags-to-push.outputs.primary-tag }}
+
+      - uses: sigstore/cosign-installer@6f9f17788090df1f26f669e9d70d6ae9567deba6 # v4.1.2
```
ldziedziul: Already installed at [L143](https://github.com/hazelcast/hazelcast-docker/pull/1301/changes/BASE..a801047e6e28ba6aeba82f94c4ca22e84d6c9c6a#diff-f23c04297854b0de26c83e33985d9ee8bbd18d70dca9291501fa34d7ec3b32cdR134) in this job - this one can be removed. Same in `ee-nlc-tag-package.yml`

.github/workflows/tag_image_promote_docker_registry.yml
```diff
@@ -88,33 +90,43 @@ jobs:
...
 
       - uses: hazelcast/docker-actions/get-jfrog-credentials@master
+        if: contains(env.REMOTE_REGISTRY, 'docker.hazelcast.com')
         id: authenticate-jfrog-write-dev-account
         with:
           aws-role-to-assume: ${{ vars.AWS_HAZELCAST_OIDC_GITHUB_ACTIONS_ROLE_ARN }}
           jfrog-oidc-provider-name: ${{ github.repository_owner }}-write-dev-account
 
       - uses: docker/login-action@v4
+        if: steps.authenticate-jfrog-write-dev-account.outcome == 'success'
```
ldziedziul, line 100: Why not the same `if: contains(env.REMOTE_REGISTRY, 'docker.hazelcast.com')` as the step above? Equivalent, just one condition to read instead of two. No strong opinion.

.github/workflows/tag_image_promote_rhel.yml
```diff
@@ -162,16 +162,37 @@ jobs:
           # Add date suffix to create unique output tag (e.g. 5.4.1-jdk17-20250909130347)
           echo "tag=${{ steps.get-tags-to-push.outputs.primary-tag }}-$(date +"%Y%m%d%H%M%S")" >> ${GITHUB_OUTPUT}
 
+      - if: steps.image-already-published.outputs.IMAGE_ALREADY_PUBLISHED != 'true'
```
ldziedziul: trailing whitespace

ldziedziul:
- Should we document how users verify the signature (basically the `cosign verify` from the description) in the README, e.g. under "Docker Images Usages"? It gets synced to Docker Hub via `update_readme.yml` too. Fine as a follow-up PR.
- Does it mean we'll be locked in with GH as the producer/signer of our images? What if we need to build the image locally to be releases (for some reason)? Will it be possible?

.github/workflows/tag_image_promote_rhel.yml
JackPGreen: fcdb5456414687b4e21f1d79282a4475b202abe5

.github/workflows/tag_image_promote_docker_registry.yml
JackPGreen: _If_ we changed the docker registry URL, we'd need to update this condition even though the step itself doesn't need to be changed. But both work and it's unlikely.

.github/workflows/tag_image_package.yml
JackPGreen: Good spot

55c9d878da59d554b46c77d228d3ef984cacfb67

.github/workflows/tag_image_package.yml
JackPGreen: Good idea

d412ec221f7324359ac0f2433815fef85cda4fa5

.github/workflows/tag_image_package.yml
JackPGreen: Good point

68b848d0f8ec6bd3ecd71cc4e8ac342fe8740c63

JackPGreen:
> Should we document how users verify the signature (basically the `cosign verify` from the description) in the README, e.g. under "Docker Images Usages"? It gets synced to Docker Hub via `update_readme.yml` too. Fine as a follow-up PR.

This is a good question and one I didn't properly explain - although you _can_ verify the signature via `cosign verify`, I would expect an end user to "notice" the signature via their own supply-chain tooling.

e.g. `docker scout` uses the _existence_ of a signature as part of it's image as _part_ of it's `docker scout quickview` scorecard.

_I think_ it's better to leave them undocumented - if a user requires such a functionality, they will know how to find it (e.g. the same applies to Maven repo checksums).

> Does it mean we'll be locked in with GH as the producer/signer of our images?
> What if we need to build the image locally to be releases (for some reason)? Will it be possible?


We won't be _locked_ but if we change where it's built, this will be observable to an end user even if the image is identical.

However, while this PR just authenticates the image was built in GitHub Actions - [GitHub Actions is implicitly SLSA L2+](https://slsa.dev/spec/v1.0-rc1/requirements), so we get that certification for "free" which isn't possible when building locally.

.github/workflows/ee-nlc-tag-package.yml
nishaatr: looks like used for actually getting the token and not for verifying signature. 
so its different. I thought it was the same endpoint but clearly not.

nishaatr:
> > I think 1 sec may be too quick for sleep? Thinking of RHEL as we know can be slow
> 
> It's polling for completion every second - not waiting one second only.

Yes. I mean 1 second for polling feels high attempt rate
But your comment `# Image signatures can take a couple of seconds to appear after the main image (suspect "oras cp --recursive" is not fully atomic), so poll for completion` does say takes couple of seconds so its fine

nishaatr (approved):

JackPGreen:
> > > I think 1 sec may be too quick for sleep? Thinking of RHEL as we know can be slow
> > 
> > 
> > It's polling for completion every second - not waiting one second only.
> 
> Yes. I mean 1 second for polling feels high attempt rate But your comment `# Image signatures can take a couple of seconds to appear after the main image (suspect "oras cp --recursive" is not fully atomic), so poll for completion` does say takes couple of seconds so its fine

Right, maybe 5 seconds is safer? It's a CI job, no-one will notice a pause 5 seconds longer than _absolutely_ necessary.

8bb8e56163334e23957064bd38c4e6865a914764

PR #1306, opened 2026-09-06
Fix non-reproducible cacerts layer in OSS image

ldziedziul:
## Summary

- `apk upgrade` re-fires the `ca-certificates` package's trigger, which regenerates `/etc/ssl/certs/java/cacerts` via the same `trust extract` command the Dockerfile already runs deterministically earlier - but the trigger-fired invocation doesn't inherit `SOURCE_DATE_EPOCH` from the build environment, so it silently falls back to wall-clock time and overwrites the deterministic file, making that layer differ on every build (caught intermittently by `verify-layer-reproducibility`).
- Fix: run `apk upgrade` *before* `ca-certificates`/`java-cacerts` are installed, while there's no trigger registered yet to fire. The single `trust extract` call afterward then stays the last write to that file - no duplicated command needed.

## Root cause

`alpine:3` has neither `ca-certificates` nor `java-cacerts` preinstalled, so nothing is registered to react to the base-package `apk upgrade`. Once `java-cacerts` is installed later (pulling in `ca-certificates`), its `update-ca-certificates` hook (`/etc/ca-certificates/update.d/java-cacerts`) fires once, and the Dockerfile's existing explicit `trust extract --overwrite ...` call immediately after it is the last word - deterministic because BuildKit injects `SOURCE_DATE_EPOCH` into that `RUN`'s environment and `trust extract` respects it there.

## Test plan

- [x] Built `hazelcast-oss/Dockerfile` twice locally with CI's exact flags (`SOURCE_DATE_EPOCH=0`, `docker-container` buildx driver, `--output type=docker,rewrite-timestamp=true`) - manifest, config, and all layer digests are now byte-identical between the two builds.
- [ ] `verify-layer-reproducibility` passes on this PR


ldziedziul:
Closing - reopening from a branch on this repo directly instead of my fork.

